PR #2, opened 2019-10-28
First attempt at fiixing identifiers

Diet-Cola:
Need to look re-look at enchantments since the ones that are erroring seem to be still correct based on https://hub.spigotmc.org/javadocs/bukkit/org/bukkit/enchantments/Enchantment.html

Please put on CivTest and reply with any errors


Maxopoly:
Thanks

PR #5, opened 2020-04-17
RandomSpawn: add city spawnpoints

Gjum:
probably missing some of the newest cities but this is a good first set


Maxopoly:
Very surprised how you whipped this out so quickly, but alright

Gjum:
I've had this lying around all finished for a while now, was just waiting on more responses (westminster, yoahtl, vinland wanted in but didn't send anything yet)

PR #8, opened 2020-05-02
Correction/additions to recipes for Smooth_Stone

crimsonblod:
Added:
-Recipes
    -smelt smooth stone
    -make smooth stone slab

Renamed/Corrected:
- Recipes that claimed they created `smooth_stone`, that actually created `stone` were retitled, and the involved "make" functions were corrected.
- Because `double_stone_slab:8` blocks no longer exist, and are now functionally the same thing as `smooth_stone`, the lore of `double block` was removed, and the output doubled, to give a equal number of smooth stone slabs as crafting by hand.
- Because `double_stone_slab:8` was essentially two smooth stone slabs, the corresponding recipe for `stone_slabs` would have just been a regular `stone` block, so the lore of `double block` was removed.

Balance:
- Gave a 1.25x multiplier for smooth stone when smelting. (64 in:80 out)
- Because the original slab recipe assumed you would be creating a double block, rather than individual half slabs, the I/O ratio was 1:1. In order to match vanilla, the output was increased to 128.

### DIFF
--- a/templates/public/plugins/FactoryMod/config.yml.j2
+++ b/templates/public/plugins/FactoryMod/config.yml.j2
@@ -68,6 +68,7 @@ factories:
         amount: 512
     recipes:
      - smelt_stone
+     - smelt_smooth_stone
      - smelt_glass
      - smelt_netherrack
      - smelt_red_netherbrick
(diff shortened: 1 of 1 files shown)


Maxopoly:
> Because the original slab recipe assumed you would be creating a double block, rather than individual half slabs, the I/O ratio was 1:1.

The ratio was actually 1:1 beforehand as well, a single slab would be replaced with a full block when being placed, but either way its fine.

Maxopoly:
Nice

PR #11, opened 2020-05-03
Charcoal Factory Corrections

crimsonblod:
Due to the change from damage values to independent blocks when defining block varients, the current state of the charcoal factory was limiting inputs to oak logs exclusively, rather than the expected behavior of using whatever log they were inputting to create charcoal.

I just made a new recipe for each log type to convert to charcoal. Unless there exists a way for factorymod to lump all logs together still, this is the best solution to the issue I could come up with.

An advantage of doing things this way though, is if people wanted to, you all could tweak individual log to charcoal values depending on say, how difficult it was to automate and/or chop down each type of tree. But that's by no means necessary.

### DIFF
--- a/templates/public/plugins/FactoryMod/config.yml.j2
+++ b/templates/public/plugins/FactoryMod/config.yml.j2
@@ -85,7 +85,12 @@ factories:
         material: CHARCOAL
         amount: 512
     recipes:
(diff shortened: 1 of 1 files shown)


Maxopoly:
Nice

PR #12, opened 2020-05-05
Removing Dirt!

crimsonblod:
- Removed line (6) `default_menu_factory: Dirt Factory`

- Converted all mentions of `dirt!` to `dirt` except for line (5948), which, as best as I can tell, has to do with fossils.

- Removed all lines made up entirely of `        name: Dirt!`

- Added a blank line at the end cause github was asking me to.

- Removed all recipes accessed via the `Dirt Factory`, including all recipes in the attached file.
[new 2.txt](https://github.com/CivClassic/AnsibleSetup/files/4578605/new.2.txt)


Maxopoly:
Merging because updated FactoryMod is on live

PR #17, opened 2020-05-06
Spawn Item Changes

crimsonblod:
- Added an `OAK_BOAT` to the a new player's spawn inventory
- Gave `OAK_BOAT` unique lore to match the the other spawn items
- Corrected `BED` to `WHITE_BED` to be consistent with 1.14's item naming, as and I chose white specifically to allow new players to dye their beds to whatever color they'd like.
- Confirmed that `COOKIE` is the correct item type.


Maxopoly:
Nice

Issue #13, opened 2020-05-05
SpawnEggs from fossils don't work


Maxopoly:
Unconfirmed, but reported

Maxopoly:
Apparently this is because of ClearLaggs limiting

Maxopoly:
Seems fixed, comment if stil an issue

PR #18, opened 2020-05-06
Attempt to fix mob egg issue

Diet-Cola:
Setting this to a higher number should help with mob egg spawns as it did locally. However if people still cant spawn mob eggs from this high a setting, might be best to disable but keep monitoring.
#13

### DIFF
--- a/templates/public/plugins/ClearLag/config.yml.j2
+++ b/templates/public/plugins/ClearLag/config.yml.j2
@@ -141,7 +141,7 @@ firespread-reducer:
 #     and possibly removed.
 chunk-entity-limiter:
   enabled: true
-  limit: 5
+  limit: 30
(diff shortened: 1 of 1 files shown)


Maxopoly:
We already have way too many mobs though? 30 per type is insane, we're already facing lag issues because of mobs.

Diet-Cola:
I think thats todo with the global cap of mobs being 2200 spread across x players, ClearLagg's default config is 300 mobs, 200 animals.

Maxopoly:
Ok, but either way this is really the wrong approach for mixing mob eggs, it's just removing the safety regulations that are rightfully keeping them from working atm.

Maxopoly:
Rejecting this for now

PR #27, opened 2020-05-18
Added recipe for cooking kelp into dried kelp

Falvyu:
This adds a recipe in the Grill to cook kelp into dried kelp.

The conversion rate is the same as other grill recipes. The grill was chosen over other factories because dried kelp is edible and because from a realistic perspective, a stone smelter would be overkill for drying up algae.


Maxopoly:
Cool

PR #31, opened 2020-05-20
Kayla HO config

Maxopoly:


Maxopoly:
@KaloHG 
So you 
- reduced gold ore rates by a factor of ~6 and made it only spawn below y 33 instead of 70
- made redstone not spawn above y 15 instead of 70, but increased its rate by a factor of 5
- made lapis ore not spawn above y 33 instead of 70, but reduced its rate by a factor of 3.75
- Scrambled all existing veins (not entirely sure how those tie into the generation rates)

Why?

The identifier fixes are fine, but this diamond issue needs to be approached from the generation algorithm, seeing what changed there from teals to dans version and how the config needs to be changed to adjust, not with arbitrary numbers thrown out there. Also whats with these severe changes to the other ores?

KaloHG:
Just went based off of the config TealNerd left in the original 1.12.2 HiddenOre Files. Those of course can be changed back.

PR #32, opened 2020-05-21
Added CustomTab+Config, Fixed HO Config.

KaloHG:
CustomTab is not present on CivClassic, and was a nice QOL improvement. So I re-added it to Ansible in hopes that it would be re-added to the main server. 

I also took a little time to take a look at the HO Config and fix it, I reverted spawn rates back to what I felt was reasonable, with the exception of the Diamond spawn rates. The prior issue with Diamond spawning was the fact that the chance to get a vein was so low, and the veins were so small and not spread out enough to allow many more diamonds to even spawn. The new config merely increases the chances to stumble upon a vein, but also makes the vein wider in turn permitting the player to actually get a decent amount of diamonds out of said vein. The generator was never the issue, the chances just somehow got broken between updates. I've tested it out myself and the new config reproduces exactly what CivClassic diamond generation was like prior.


Maxopoly:
Try to separate different changes into separate PRs, otherwise any single problem with your PR is gonna prevent it from getting merged entirely

- You still have various changes to other ore rates in this PR, is that intended? Either way I'm opposed to those

- Regarding the diamond rates, please reread what I wrote on them [here](https://github.com/CivClassic/AnsibleSetup/pull/31#issuecomment-631680507) . [Here](https://pastebin.com/wqK0UYqa) is the previous 1.12 config for reference.

- If you change serialized item stacks from their legacy identifiers, you need to add `v: 1` to the entry

PR #36, opened 2020-05-27
Changed recipes to use the new white/brown/blue/black dyes

Falvyu:
The new dyes can be crafted using the old materials at a 1:1 ratio. 
Unlike the older ones, they can also be crafted with the new flowers, hence the PR.


Maxopoly:
Nice

PR #39, opened 2020-06-11
Update config.yml.j2

Protonull:
Fixes [RealisticBiomes issue 8](https://github.com/CivClassic/RealisticBiomes/issues/8)


Protonull:
When Realistic Biomes parses its config, it prioritises Biomes over Biome Aliases, thus when it sees "forest" or "swamp" it thinks of the _biome_ and not the alias, so the other forest-like and swap-like biomes don't receive any growth modifiers.

Maxopoly:
Good detective work

PR #40, opened 2020-06-16
Nerf bastions further

Xavter:
Make pearls destroy bastions in an hour vs slightly over that, make block placements destroy mature bastions in 45 minutes vs 60 minutes, make immature bastions break in 5mins v 7mins.

### DIFF
--- a/templates/public/plugins/Bastion/config.yml.j2
+++ b/templates/public/plugins/Bastion/config.yml.j2
@@ -26,8 +26,8 @@ bastions:
     squarefield: true
     effectRadius: 10
     includeY: true
-    startScaleFactor: 10
-    finalScaleFactor: 1.111111111111111
+    startScaleFactor: 13.33333333333333
+    finalScaleFactor: 1.5
     warmupTime: 172800000
     erosionPerDay: 0
     regenPerDay: 5
@@ -42,7 +42,7 @@ bastions:
     pearls:
       block: true
       blockMidair: false
-      scaleFactor: 1
+      scaleFactor: 1.111111111111111
       requireMaturity: true
       consumeOnBlock: false
       damageFirstBastion: true


Xavter:
I believe supervaults are still very powerful and we need to add buffs to both pearls and block placement erosion.

Maxopoly:
For now I do not think that this is the right approach to addressing the issues we are seeing in regards to Bastions.

Xavter:
Can you clarify what the alternative approach would be? Vaults at this stage are extremely overpowered and there is no incentive to attempt breaks when a >20 ring vault takes several days of breaking

Maxopoly:
I think breaking vaults is difficult, but not impossible. The overlap damage change already was a heavy nerf that went without consequences, to me it seems like the main thing currently keeping players from breaking bigger vaults is a lack of motivation to play in general, not vaults being unbreakable. Non-mega/normal vaults were already weakened significantly and I do not want to further push that and make everything but ~4 vaults on the map de facto useless.

I think with CivRealms as an alternative and the map being kinda stale a significant part of it is a motivational issue, we saw the same during the Lexington war. I am arrogant enough to believe that if I were to still play, I could again lead WP to victory and there are other people who could do so as well, but just don't care enough to do so.

Disregarding your passive aggressive attitude, I do not think that this is the single key to removing Classics' chains, that view is too focused on a specific groups of players. Even if the current conflict was entirely gone, only a very small portion of the players choosing CivRealms instead would have their situation significantly changed, most of them are bound by bans, perma pearls or just the staleness of an older map in general.

Xavter:
If you caught notes of passive aggressiveness when there was none in the writing, then I'm afraid this discussion will devolve into accusations of personal attacks at each turn. I suggest for a sane discussion you leave design decisions open to criticism without attaching harshness towards your character or the tone of my content.

I agree that it's theoretically possible to break even 100-ring vaults, but I think almost every current player who have been breaking vaults would agree even a force led by you would be able to break vaults in their current state, even with the force we had gathered at Lexington. The Lexington vault had an overwhelming numerical advantage that shouldn't have been necessary, while having a fraction of the current rings of the largest vaults. Ignoring any personal judgements about "player motivation", if you look at the perspective of hours alone required for the break of pinkerton, the current vault meta is significantly more difficult than the Lexington break. The time it took to bring down Pinkerton is proof of the current meta being broken, for example.

Regarding balancing around "specific players" or around 4 vaults: yes, we are balancing around the overpowered. If you allow an economy where megavaults can be realistically attained, balance needs to revolve around the megavaults primarily, otherwise you've introduced a game breaking mechanic. In EVE or Overwatch, balance changes have prioritized removing too overpowered mechanics, because overpowered mechanics exclusively favor the people doing the gamebreaking. 

Should there be attention paid to balance smaller vaults? Absolutely, but the priority in the immediate term should be to end overpowered mechanics, otherwise we have a situation where 40 men in prot can't topple a 30 ring vault defended by 5 people unless they spend 4 continuous days attacking it. We may need different mechanics entirely to bring smaller vaults up to par with megavaults, but even if that's the case we shouldn't wait for someone to code it themselves to fix the more pressing gamebreaking issue.

If any current megavault owners wanted to pearl or obby bomb the remaining players on classics, they could do so with impunity, and there would be nothing capable of stopping them - it doesn't take much of a surplus of mana to harass a player long enough to cause them to quit. 

Impossibility to enforce consequences on megavault owners is why I brought up the point of server death: should we endorse it at this point if the idea that balance is currently attained is being based on your personal judgements of player motivation? I am favoring that outcome if we decide the status quo is correct, and I think you should fairly communicate that you disagree with making changes to add in-game enforceable consequences for megavault owners, which would indicate for most the end of the server.

crimsonblod:
I know I'm not a vault owner, but I think it's important to actually see how players handle breaking a vault with the current changes before making more of them IMO. Whether any new clever strategies are invented and such that make it easier to break vaults than the original changes intended, etc... I think changing too much at once for vaults in particular, may cause more harm than good. At least too much to quickly rather. Although talking to whoever set up that pvp server with practice vault attacks may be a good idea if you want to accelerate testing like that, to see how they did it, so potentially the test server could do the same.

Protonull:
Another non vault owner (or pvper) chipping in here :P one idea I've had is slightly expanding the world border by 100-200 blocks, and having that new area be unreinforceable. This wouldn't be done immediately, rather an announcement would be made and a couple of months notice given for nations with world border infrastructure to make plans. Just an idea :)

PR #46, opened 2020-07-07
Update configs

Protonull:


templates/public/plugins/RailSwitch/config.yml.j2
```diff
@@ -1,13 +0,0 @@
-# can be mysql or sqlite
```
Maxopoly, line 1: Does RS really not use a config at all anymore? Even if it doesn't I'd prefer keeping an empty config here.

Protonull: Nope, no config anymore since it doesn't use databases anymore but player settings

PR #49, opened 2020-07-13
TradePlus config changes

Protonull:


templates/public/plugins/TradePlus/config.yml.j2
```diff
@@ -13,7 +13,8 @@ permissions:
 requestcooldownseconds: 10
 allow-trade-in-creative: false
 blocked:
-  blacklist: []   
+  blacklist:
+  - ender_pearl
```
Maxopoly, line 17: Could we instead use the lore specific to pearls so normal ender pearls can be traded?

They will always have Light green color + `Exile Pearl` or `Prison Pearl`

Maxopoly:
Neat

PR #50, opened 2020-07-14
Orebfuscator update

Protonull:
The overall changes will just show all green. Look at the individual commits to see what's happening.

I've added back the missing entries from the [old config](https://github.com/CivClassic/Configs/blob/master/Orebfuscator4/config.yml#L57-L59), as well as added the new blocks from 1.13 and 1.14. I've left out the blocks that are exclusively villager task based (such as Smithing Tables) and blocks that have aesthetic purposes (like Campfires and Lecterns). Other blocks that are disabled in Civclassic (like Scaffolding and Grindstones) have been left out.

Some of the changes to `RandomBlocks` and `ObfuscateBlocks` are just re-organisation to put the ores together and such so it's more readable for future contributors. I've removed ore completely as the.. confetti blocks.. for a lack of a better term, since people seem to fine them frustrating when mining. TNT has been removed due to obvious issues with world downloads.

### DIFF
--- a/templates/public/plugins/Orebfuscator/config.yml.j2
+++ b/templates/public/plugins/Orebfuscator/config.yml.j2
@@ -1,0 +1,116 @@
+ConfigVersion: 13
+Booleans:
+  UseCache: true
+  Enabled: true
+  UpdateOnDamage: true
+  NoObfuscationForMetadata: true
+  NoObfuscationForOps: false
+  NoObfuscationForPermission: false
+  LoginNotification: true
+Integers:
+  MaxLoadedCacheFiles: 64
+  DeleteCacheFilesAfterDays: 0
+  EngineMode: 2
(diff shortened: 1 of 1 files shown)


Maxopoly:
Unrelated to your PR, but this config format of dividing up based on type of parameter is very strange. Is it possible to do per biome overrides for the nether? With this setup any snitches etc. would still stick out in all the netherrack.

Does this also randomly obfuscate occluded stone?

Protonull:
[Here's](https://github.com/lishid/Orebfuscator/wiki/Configuration-File) the config file wiki. It seems to be based on world type but also seems to gracefully handle nether biomes inside default world types... or appears to based on the default config that Orebfuscator was using prior to this explicit config.

----

> Does this also randomly obfuscate occluded stone?

No idea

PR #53, opened 2020-07-15
hint hint

Protonull:


Maxopoly:
The MOTD in the bungee config is the only one that matters, but yeah

PR #55, opened 2020-07-22
Fix fire invulnerability ticks

jaymsDooku:
As much fun as [this is](https://streamable.com/ux0vmr), let's fix that.


Maxopoly:
Thanks

PR #56, opened 2020-07-22
Update config.yml.j2

crimsonblod:
- Changed bakery construction and repair to individual `BRICK` from `BRICKS` (The block)
(Construction may be reverted back of course, but the couple people I talked to remember it being individual bricks, not brick blocks)

- Changed coral block recipes from using `GRASS` to `GRASS_BLOCK`


Maxopoly:
Cool

Issue #29, opened 2020-05-19
Heart of the Sea attainability/ideas for fossils


crimsonblod:
Should we add hearts of the sea to something like say, fossils?

It appears that the only vanilla way to get them is [via chests that are created at worldgen](https://minecraft.gamepedia.com/Heart_of_the_Sea). And without it, it seems conduits are essentially disabled. Before implementing a change like this, we should test to see if conduits break anything with the plugins in their current states.

A factory recipe could also be good, but IMO, fossils are a better route to give another useful item based reward to their prize pool.

Along that line of thought, a skunkworks level idea could be to add a fairly low chance of getting mana from fossils as well?

KaloHG:
Hi crimson! Looking through what conduits do to the player via effects and other various things, it appears they would not affect any plugins currently ran on classics. However, some issues with Citadel could arise with reinforcing conduits. As the game treats conduits as entities similar to End Crystals. 

For a fossil drop rate. I'd reccomend making it semi-common, probably below heads. 

Conduits will likely become an extremely important part of gameplay. I can see their use ranging from simple city assistance. To vault construction/protection. However, currently since conduits are not obtainable to gameplay, we have not seen the potential effects. 

I'll see what I can do to test out conduits on an ansible setup server. Just to see if it breaks anything.

KaloHG:
Tested it out. All seems well, Citadel works fine, etc. No bugs noticeable initially. It is rather cheap to make though. Maybe making Heart of the Sea a factory recipe of the sorts instead?

Maxopoly:
Addressed by https://github.com/CivClassic/AnsibleSetup/pull/57

PR #57, opened 2020-07-23
Replace crack_fossil bucket output with heart of the sea

Inframission:
Made in response to https://github.com/CivClassic/AnsibleSetup/issues/29 
Unsure whether fossil 'chance' values need to be specific so I've replaced the bucket output


Maxopoly:
> Unsure whether fossil 'chance' values need to be specific so I've replaced the bucket output

The sum should be 1.0, so yeah

PR #59, opened 2020-07-26
Enable block limits

squareblob:
- Currently CivClassic seems to have nothing enforcing block limits (eg. try placing 64 dropper's in a single chunk). 
- Remove SLIME_BLOCK as SAH only limits tile entities.


Maxopoly:
Nice

PR #63, opened 2020-07-28
Fixed wrong obfuscation material name

Falvyu:
End stone was registered as ENDER_STONE rather than END_STONE which prevented plugin from functioning properly.


Maxopoly:
Superceded by https://github.com/CivClassic/AnsibleSetup/commit/81f7fe8845b132e3e91c2b3a2a6410f698c8a966 because fixing the plugin naming (and thus renaming the config) merge conflicted this PR. Also you missed one instance of ENDER_STONE.

PR #65, opened 2020-07-28
Scouter9001 - Grow mushrooms in bio lab factory.

Scouter9001:
Enable players to create brown and red mushrooms in the bio lab.

1.  Automatic mushroom farms can cause server lag
2.  AFK'ing a mushroom farm is tedious, encourages ALTs, and keeps chunks loaded
3.  The bio lab already has recipes to make mushroom blocks from mushrooms.  This enables a natural progression from creating mushrooms, to creating mushroom blocks from mushrooms.
4.  Due to non-linear spread mechanics, persisting mushroom growth could cause additional resource utilization from the server to track things like cross-chunk spreading and loading.

I've created an initial recipe that is open for discussion.  I liked the idea of combining "gross things" to simulate the soil environment that a mushroom would naturally appear in.

### DIFF
--- a/templates/public/plugins/FactoryMod/config.yml.j2
+++ b/templates/public/plugins/FactoryMod/config.yml.j2
@@ -379,6 +379,8 @@ factories:
      - acacia_to_sapling
      - sapling_to_dark_oak
      - dark_oak_to_sapling
+     - grow_mushroom_brown
+     - grow_mushroom_red
      - make_mushroom_brown
      - make_mushroom_red
      - make_mushroom_stem
@@ -4834,6 +4836,50 @@ recipes:
         amount: 128
         lore:
           - Crate
+  grow_mushroom_brown:
+    type: PRODUCTION
+    name: Make Brown Mushrooms
+    production_time: 60s
+    fuel_consumption_intervall: 2s
+    input:
+      podzol:
+        material: PODZOL
+        amount: 64
+      spider_eye:
+        material: SPIDER_EYE
+        amount: 128
+      red_dye:
+        material: BROWN_DYE
+        amount: 64
+      white_dye:
+        material: WHITE_DYE
+        amount: 64
+    output:
+      brown_mushroom:
+        material: BROWN_MUSHROOM
+        amount: 64
+  grow_mushroom_red:
+    type: PRODUCTION
+    name: Make Red Mushrooms
+    production_time: 60s
+    fuel_consumption_intervall: 2s
+    input:
+      podzol:
+        material: PODZOL
+        amount: 64
+      spider_eye:
+        material: SPIDER_EYE
+        amount: 128
+      red_dye:
+        material: RED_DYE
+        amount: 64
+      white_dye:
+        material: WHITE_DYE
+        amount: 64
+    output:
+      red_mushroom:
+        material: RED_MUSHROOM
+        amount: 64
   make_mushroom_brown:
     type: PRODUCTION
     name: Make Brown Mushroom Blocks
--- a/templates/public/plugins/FactoryMod/config.yml.j2
+++ b/templates/public/plugins/FactoryMod/config.yml.j2
@@ -4839,7 +4839,7 @@ recipes:
   grow_mushroom_brown:
     type: PRODUCTION
     name: Make Brown Mushrooms
-    production_time: 60s
+    production_time: 10s
     fuel_consumption_intervall: 2s
     input:
       podzol:
@@ -4861,7 +4861,7 @@ recipes:
   grow_mushroom_red:
     type: PRODUCTION
     name: Make Red Mushrooms
-    production_time: 60s
+    production_time: 10s
     fuel_consumption_intervall: 2s
     input:
       podzol:
--- a/templates/public/plugins/FactoryMod/config.yml.j2
+++ b/templates/public/plugins/FactoryMod/config.yml.j2
@@ -4848,6 +4848,9 @@ recipes:
       spider_eye:
         material: SPIDER_EYE
         amount: 128
+      phantom_membrane:
+        material: PHANTOM_MEMBRANE
+        amount: 16
       red_dye:
         material: BROWN_DYE
         amount: 64
@@ -4870,6 +4873,9 @@ recipes:
       spider_eye:
         material: SPIDER_EYE
         amount: 128
+      phantom_membrane:
+        material: PHANTOM_MEMBRANE
+        amount: 16
       red_dye:
         material: RED_DYE
         amount: 64


Maxopoly:
Curious to hear some other peoples thoughts on this, I have no clue whether this is a good idea or not.

wingzero54:
The new recipes call for 336 blocks (5 different block types) to create 64 mushrooms. I think that is too steep of an exchange. It calls for 128 dye total, I think it would be fine to only call for 16 dye total.

As for balance, mushrooms aren't really considered a bottleneck and this recipe is not too cheap. I think we need to hear from more people though

Scouter9001:
Completely open to adjusting down.  Didn't want it to be too easy to try and match the difficulties encountered growing them due to slow spread and chunk loading.  

What about:

- 64 Podzol
- 64 Spider Eyes
- 16 Phantom Membranes
- 16 Brown Dye
- 16 White Dye

Xavter:
I'm opposed to factory recipes to replace in-game growth mechanics in general. It can replace interesting pieces of infrastructure with our less interesting factory boxes. 

However, I agree the issues Scouter brings up regarding current mushroom farm infrastructure are undeniable and demonstrate one of our shittiest decisions to include them as XP ingredients.

I think there's a middle ground that takes more work, but results in interesting mechanics by avoiding mushroom printing presses: randomly generated mushroom plants using mushroom blocks and the L-system tree generators @Maxopoly was working on as part of [RB issue 7](https://github.com/CivClassic/RealisticBiomes/issues/7) on [avocado](https://github.com/CivClassic/RealisticBiomes/tree/avocados). This way we could resolve mushroom farms creating heavy load and also make further progress on/try out a very interesting mechanic at the same time.

Curious what your thoughts are @Scouter9001 and @Maxopoly as I believe significant work was done on the L-system trees even though they were not completed.

Scouter9001:
Thanks @Xavter!  Yeah I want to make sure we don't enter a slippery slope of replacing everyday things with a grind it out recipe in factory mod. 

**General Thoughts:**

Large mushroom trees seem to be disabled from growing out of mushrooms, is that accurate?
I would love to see mushroom trees sprout up randomly.

I do think factory mod would be the simplest to implement as it would take a fair amount of time and testing in order to develop a time boxed persistence system that would handle spread mechanics when 1 mushroom could cause the spread to hop back and forth between chunks.  

Also, I think it would take a some small to medium level of effort and testing to force spread mechanics into the same chunk the "prime" mushroom was planted in.  Testing could be sped up by upping the randomTickSpeed of course.  This latter approach would make farming more difficult for the user however to account for containing mushroom growth in the same chunk and maintaining vanilla spread mechanics of 5 of a kind within X blocks.  But hey, "farming ain't easy."

L-System Trees:  Super cool.  If we can have x% chance a mushroom spawns a tree that would be great as long as the % chance is sane enough.  Another riff on the idea is that a certain block or block combination increases that growth chance such as the clay block mechanic.

**Side notes:**
Beyond technical reasons... growing things in a bio lab is pretty cool.  Just like a local makerspace that has one - you can create/grow all sorts of things.

On a side note: Without knowing the full context, I'd love for a composter to be apart of the recipe in a factory mod based approach hah.

Scouter9001:
@Maxopoly Base branch changed to civ16

Nuclear-Crow:
I don't like this idea. It removes one of the farms necessary for XP production, which is not a good thing. You could just remove shrooms from the XP recipe in this case. Shroom trees sound like a nice solution though.

crimsonblod:
I do like the idea of using membranes in some sort of farming application though. Random ideas come to mind like a version of bonemeal that works on crops, or a reinforcement required to make clay work or something.

Falvyu:
@Scouter9001 I don't know if you have checked the subreddit but crimsonblod made a [post](https://www.reddit.com/r/civclassics/comments/hzhmzi/discussion_request_pr65_major_mushroom_farmingxp/) about this PR a few days ago.

Anyway, Xavter advised me to post my comment here as well so here it is:

__[First comment](https://www.reddit.com/r/civclassics/comments/hzhmzi/discussion_request_pr65_major_mushroom_farmingxp/fzj7006/)__
I'm a bit skeptical about this change.

> Automatic mushroom farms can cause server lag
> Due to non-linear spread mechanics, persisting mushroom growth could cause additional resource utilization from the server to track things like cross-chunk spreading and loading.

How does this compare to other afkable farms (e.g: portal farms, spider grinders, flower farms, sugar cane/melon farms) ?

How much impact do these farms (and other RB-based farms) have on server performance (ie: tps) right now ?

Is this the only solution to reduce the amount of lag generated by mushrooms farm ? "non-linear spread mechanics" refers to the fact that mushrooms spread randomly around every <X> ticks rather than instantly growing when loading the chunk, right ? If so, would it be realistically possible to change this behavior to a "linear spread", e.g: when loading the chunks, the mushrooms will try to spread at a very high rate or something like that ?

> AFK'ing a mushroom farm is tedious, encourages ALTs, and keeps chunks loaded

While this statement isn't wrong, I'm a bit skeptical on whether mushrooms farms are the real issue here.

AFAIK mushrooms grow (or used to) at the same rate everywhere and grow underground. If that's still the case then you can always set up a farm near a town to have it frequently loaded. It's obviously not the perfect solution if you're regularly harvesting it with water because it'll make fps' drop but we have something like that in the Commonwealth and it hasn't really been an issue. In the past, I've also worked on XP runs when I was in Vinland. Mushrooms weren't the bottleneck and afking it for a few hours was enough for an advanced run. This means that most of the time spent afking for XP production wasn't spent on mushrooms. Some people might have been afking mushrooms 24/7 but it's very likely that they also automated their crate production + the rest of the farms to back it up.

> What about: (per stack) 64 Podzol 64 Spider Eyes 16 Phantom Membranes 16 Brown Dye or 16 Red Dye 16 White Dye

This would be okay as an alternative way to get a small amount of mushrooms. However, this is extremely expensive if you're trying to replace mushrooms farms. Dividing the cost by 4 might be more balanced. Still, I do like the idea of introduce new materials to grind like membranes.

And another thing: Is it a good idea from a gameplay perspective ? There are people out there who enjoy designing farms. Is replacing it with a factory recipe really more interesting than that gameplay wise ?

__[Second comment, in response to Xavter's remark up here](https://www.reddit.com/r/civclassics/comments/hzhmzi/discussion_request_pr65_major_mushroom_farmingxp/fzkg37o/)__

> I'm opposed to factory recipes to replace in-game growth mechanics in general. It can replace interesting pieces of infrastructure with our less interesting factory boxes.

I completely agree with that and that's my main issue with the PR.

> However, I agree the issues Scouter brings up regarding current mushroom farm infrastructure are undeniable and demonstrate a poor decision to include them as XP ingredients. 

If this refers to the lag issues then I think it'd be nice to evaluate the impact to see if it's really that bad. It's probably possible to have someone load the farms and measure how much lag does it generates.

>  I also like the idea of bringing more ingredients into mushroom production and crop production in general.

Same. It shouldn't be too expansive but "widening the tech tree" is IMHO a good thing. Alternatively, use new materials/mechanics to give more depth to farming.
A few random ideas:

- Using the planned "fruits" or other materials (e.g: honey) to give a boost to crops in an area for a limited amount of time.

-  Give bonuses (growth rate, yield) to crops if they're adjacent to certain types of blocks/plants.

> I disagree that alts are an issue here: any non-manual farm encourages alts, we should address alt balance and automation balance holistically in a separate issue as they are currently part of balance decisions of many other config values.

I agree with you. 

> I think there's a more interesting middle ground that requires more work but avoids mushroom printing presses: randomly generated mushroom plants using mushroom blocks and the L-system tree generators @Maxopoly was working on as part of RB issue 7 on avocado trees. This way we could resolve mushroom farms creating heavy load and also make further progress on/try out a very interesting mechanic at the same time.

I love the randomly generated plants idea with generators. I'm not sure how it'd work out for mushrooms though.

__A few additional remarks__

More comments have been posted here since when I made the aforementioned comments on reddit. I'll give my opinion of them here:

> Large mushroom trees seem to be disabled from growing out of mushrooms, is that accurate?
> I would love to see mushroom trees sprout up randomly.

Giant mushrooms grow using bone meal which is disabled by [Realistic Biome](https://github.com/CivClassic/RealisticBiomes/blob/master/src/main/java/com/untamedears/realisticbiomes/listener/PlantListener.java#L67)
Maybe an exception could be made for giant mushrooms + the nether fungi ?

> Beyond technical reasons... growing things in a bio lab is pretty cool. Just like a local makerspace that has one - you can 
> create/grow all sorts of things. Although mushrooms are the only thing I'd like to see "grown" at the moment.

One major issue that I have with using Factories is that they are extremely boring from a gameplay perspective. I don't think it's necessarily a bad thing to use them to simplify tasks that are boring to begin with (e.g: waiting for a furnace to smelt stone/food/...). However, making farms isn't what I'd consider boring. Furthermore, mushrooms are XP materials and XP production should require an infrastructure to back it up.

> On a side note: Without knowing the full context, I'd love for a composter to be apart of the recipe in a factory mod based approach hah.

Do you mean that composters could be used to make "alternative" factories for these new recipes ? (e.g: a bio lab but it requires a composter block in place of the crafting table)
If so then while that wouldn't really change my view about the recipes themselves, I think it could be an interesting emergent mechanic that could bring more depth to factories.

Scouter9001:
Added a quick reply to the reddit post:
https://www.reddit.com/r/civclassics/comments/hzhmzi/discussion_request_pr65_major_mushroom_farmingxp/g046k20?utm_source=share&utm_medium=web2x

Scouter9001:
Mushroom persistence if you trace through configs from CivEx First Light:
https://github.com/CivilizatonExperiment/CivEx--First-Light-Configs/blob/master/RealisticsBiomes/config.yml

Maxopoly:
Ah, deleting the civ16 branch this was against autoclosed this PR.

Maxopoly:
I am also against replacing ingame farming with a factory, but this recipe is expensive enough to have it only serve as mean to gain a few initial mushrooms for a farm, mass production with it is not feasible (which is good). I'd be in favor of reducing the cost to what you suggested :

```
    64 Podzol
    64 Spider Eyes
    16 Phantom Membranes
    16 Brown Dye or 16 Red Dye
    16 White Dye
```

PR #74, opened 2020-08-06
Added recipe for crying obsidian

Falvyu:
This is a proposition for:

2 ghast tears + 64 obsidian => 64 crying obsidian

Crying obsidian has the same hardness as regular obsidian. The main difference is that it emits light (10).
I think the 2 ghast tears justify this small improvement, while also giving more use to the tears. 

EDIT: It is worth noting that crying obsidian can be used to make respawn anchors. They don't work in the overworld/end and their explosions are [disabled by default](https://github.com/CivClassic/SimpleAdminHacks/pull/23).

### DIFF
--- a/templates/public/plugins/FactoryMod/config.yml.j2
+++ b/templates/public/plugins/FactoryMod/config.yml.j2
@@ -4913,14 +4913,17 @@ recipes:
     input:
       obsidian:
         material: OBSIDIAN
-        amount: 64
+        amount: 128
       tears:
         material: GHAST_TEAR
-        amount: 2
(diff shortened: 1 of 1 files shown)


Maxopoly:
Seems kinda expensive. I was thinking about just using glowstone dust, because that kinda makes sense with the glow, is a nether material and pretty common as well.

wingzero54:
I think 1 ghast tear or some glowstone would both be good recipes for this.

Maxopoly:
@wingzero54 Is this fine? If yes, merge it.

PR #75, opened 2020-08-07
gray local chat for 800m+ distance (outer 200m ring)

Gjum:
this makes triangulation (using local chat colors) less effective


Maxopoly:
Need testing

PR #80, opened 2020-08-10
Add BuffSpanker support

Protonull:


templates/public/plugins/SimpleAdminHacks/config.yml.j2
```diff
@@ -221,8 +221,15 @@ hacks:
     enabled: true
   ItemMetaConverterHack:
     enabled: true
-  ItemMetaConverterHack:
```
Maxopoly, line 224: we still need this

Maxopoly: ah wait nevermind

PR #88, opened 2020-08-19
Add chains and lanterns to the iron forge

wingzero54:


templates/public/plugins/FactoryMod/config.yml.j2
```diff
@@ -3375,6 +3376,21 @@ recipes:
...
         amount: 256
+  make_chains:
+    production_time: 4s
+    name: Make Chains
+    type: PRODUCTION
+    input:
+      iron:
+        material: IRON_INGOT
+        amount: 16
+      iron:
```
Maxopoly, line 3387: I'd rather have a bit of additional iron ingot cost than including nuggets, crafting those is just annoying

Maxopoly: Also both sections under input have the same identifier, which probably isn't gonna work

wingzero54: Roger that. I also realized Lanterns would be a good addition as they're primarily made of iron, I'm going to include those as well.

Maxopoly:
Nice

PR #94, opened 2020-10-11
Tweaked fire damage tick to prevent lava instadeath

wingzero54:


Maxopoly:
This will make it so you are invulnerable to _any_ damage for 10 ticks after getting damaged by fire, making it very advantageous to be on fire in pvp. The invul system added by jayms is in general kinda non-ideal, we'd need it to do cooldowns specific to that type of damage only. Not sure how that works in vanilla. Either way this PR will break it even more

wingzero54:
Half a second (roughly 10 ticks) is the vanilla behavior for invulnerability. Either way this invul system seems like it needs more tweaking and testing. Would you prefer to just remove the fire-based invul changes for now?

wingzero54:
I've updated the commit, here is my comment:
Vanilla behavior is 10 ticks of invulnerability after any damage (excess damage is still applied - ex: you get 5dmg, then hit with 7dmg within the 10 ticks. You take the original 5dmg, plus 2dmg) https://minecraft.gamepedia.com/Damage#Immunity. Damage in Minecraft is balanced around the mechanic of the 10 tick invulnerability. Lava adds 300 fire ticks immediately https://minecraft.gamepedia.com/Lava#Burning and without this 10 tick invulnerability, the damage gets applied every single tick instead of once every 10 ticks.

So my recommendation is update fire_tick to 8 tick invulnerability (20% decrease from vanilla). The goal of this invulnerability system is to allow players to deal more damage while PvP'ing, in comparison to vanilla. This needs to be balanced with the PvE damage. This would still allow more damage to be done in PvP (in a 60 second fight you could potentially apply 30 more ticks of fire damage) while maintaining a similar PvE experience. 

I have tested this on my local server and verified it acts as intended.

KaloHG:
The current PvP kit is already pretty squishy, I think it'd be better if it's more tanky. Just purely because you can already take quite a bit of damage in a short amount of time.

wingzero54:
The current is only a 1 tick invul period for fire tick, so this would reduce the amount of damage you can currently deal in PvP (it would be closer to what it was before the Finale invul addition, in terms of fire tick damage)

Maxopoly:
I think Finales Invul system doesnt work like that though, it doesnt apply the exceeding damage afaik.

wingzero54:
Whether or not it applies exceeding damage doesn't really matter in this case. In full prot you currently die in lava in under 4 seconds, the current config is absolutely unacceptable. The tiny change in potential PvP damage dealt is immaterial compared to the way it absolutely fucks PvE.

wingzero54:
I tested pvp with Fire Tick 1 and 10. They gave identical kill times in pvp with an alt. I could not distinguish any difference in the amount of damage dealt between the two settings. So I think the pvp concerns with this change are unfounded and it should be merged.

PR #96, opened 2020-10-12
Add various spawn eggs

wingzero54:
Added Fox, Cat, Hoglin, Strider spawn eggs and fixed a few spawn egg formatting typos.

### DIFF
--- a/templates/public/plugins/FactoryMod/config.yml.j2
+++ b/templates/public/plugins/FactoryMod/config.yml.j2
@@ -5546,23 +5546,43 @@ recipes:
           amount: 1
       bee:
         chance: 0.00029411764
-        wolf:
+        bee:
           material: BEE_SPAWN_EGG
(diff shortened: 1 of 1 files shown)


templates/public/plugins/FactoryMod/config.yml.j2
```diff
@@ -5546,24 +5546,44 @@ recipes:
...
           material: DOLPHIN_SPAWN_EGG
           amount: 1
       stray:
         chance: 0.00029411764
-        wolf:
+        stray:
           material: STRAY_SPAWN_EGG
           amount: 1
+      strider:
+        chance: 0.00029411764
```
Maxopoly, line 5568: You'll need to lower some other chance, it needs to sum up to 1

wingzero54:
I've updated the chances as requested so they sum to 1

PR #101, opened 2020-10-26
Add recipes for crimson & warped fungus

wingzero54:


Maxopoly:
I kinda wanted to reserve those for RB usage but we can also just add them like this :shrug:

wingzero54:
Well there's 8 unique blocks and 24 variants from huge fungus, and huge fungus can only be grown from bonemealing the fungus.  My goal is to implement those through this. I don't think these recipes are cheap enough for large scale production, is that your concern? Huge fungus also cannot be broken down into fungus either, so it wouldn't pose a future XP concern like giant mushrooms do.

wingzero54:
Bumping this for review

PR #104, opened 2020-11-06
Add TAB plugin for custom tablist

squareblob:
[TAB](https://github.com/NEZNAMY/TAB) is licensed under Apache 2.0. Only using it for a custom header and footer, disabled all the extraneous config features.

CC


Maxopoly:
We don't do the colored name thing anymore, because noone active has access to anything related to finances. I've even kind of recommended against donating to Civclassics as I have no insight into where that money really goes

Maxopoly:
Nice, ty

PR #110, opened 2020-11-21
Add a grinding mill factory

wingzero54:
Adds some nice QOL recipes that give more lateral accessibility to get materials

### DIFF
--- a/templates/public/plugins/FactoryMod/config.yml.j2
+++ b/templates/public/plugins/FactoryMod/config.yml.j2
@@ -5027,7 +5027,7 @@ recipes:
         material: SNOW_BLOCK
         amount: 64
   crush_stone:
-    production_time: 4s
+    production_time: 16s
     name: Crush Stone
     type: PRODUCTION
     input:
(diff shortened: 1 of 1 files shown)


templates/public/plugins/FactoryMod/config.yml.j2
```diff
@@ -4979,6 +5002,102 @@ recipes:
...
+    type: PRODUCTION
+    input:
+      ice:
+        material: ICE
+        amount: 64
+    output:
+      snow:
+        material: SNOW_BLOCK
+        amount: 64
+  crush_stone:
```
Maxopoly, line 5029: Putting this in a hopper loop with a stone smelter and some delay prints stone with very slight charcoal consumption. Need to consider the same for all other recipes

wingzero54: 12 charcoal to turn 64 stone -> 96 stone, I hadn't considered that. I don't think there are any other exploits like that - sand to sandstone is 1:1 and dirt to dirt varieties is 1:1 as well. We could double or triple the charcoal for stone:cobble recipe specifically

wingzero54: I 4x the coal cost for the stone recipe, so it's 32s for each stack of stone which I think is sufficient to dissuade stone printing

PR #115, opened 2020-11-29
Re-enable smaller amount of mobs

wingzero54:


Maxopoly:
Let's give it a try. Note that the tick-per spawn interval is 2000x higher than vanilla.

PR #117, opened 2020-11-30
disable phantoms

TheOrangeWizard:
may need discussion. however phantoms are extremely unpopular and have limited gameplay value

### DIFF
--- a/templates/public/plugins/SimpleAdminHacks/config.yml.j2
+++ b/templates/public/plugins/SimpleAdminHacks/config.yml.j2
@@ -202,16 +202,16 @@ hacks:
     breakDenyDuration: 3000
     laggLenciency: 1.5
   PhantomMenace:
-    enabled: true
+    enabled: false
     # Players have a statistic called TIME_SINCE_REST. If a player's value is above 72k (3 days minecraft time) then
     # they will start spawning phantoms. Setting the cap to below 72k will effectively disable phantoms. Setting the
     # cap to -1 will disable the cap. Default: 80k (10% chance per two minutes during the night and thunder storms.)
     # Read more here: https://minecraft.gamepedia.com/Phantom#Java_Edition
     timeSinceRestCap: 75000
     # Determines whether phantoms are allowed to spawn during the night.
-    nightSpawn: true
+    nightSpawn: false
     # Determines whether phantoms are allowed to spawn during thunder storms.
-    stormSpawn: true
+    stormSpawn: false
     # Determines the maximum light level at the spawning entity's location that phantoms are allowed to spawn at.
     # Note: 15 will enable spawning at all light levels.
     maximumLightSpawn: 7
--- a/templates/public/plugins/SimpleAdminHacks/config.yml.j2
+++ b/templates/public/plugins/SimpleAdminHacks/config.yml.j2
@@ -207,7 +207,7 @@ hacks:
     # they will start spawning phantoms. Setting the cap to below 72k will effectively disable phantoms. Setting the
     # cap to -1 will disable the cap. Default: 80k (10% chance per two minutes during the night and thunder storms.)
     # Read more here: https://minecraft.gamepedia.com/Phantom#Java_Edition
-    timeSinceRestCap: 75000
+    timeSinceRestCap: 0
     # Determines whether phantoms are allowed to spawn during the night.
     nightSpawn: false
     # Determines whether phantoms are allowed to spawn during thunder storms.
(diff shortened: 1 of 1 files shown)


templates/public/plugins/SimpleAdminHacks/config.yml.j2
```diff
@@ -202,16 +202,16 @@ hacks:
     breakDenyDuration: 3000
     laggLenciency: 1.5
   PhantomMenace:
-    enabled: true
+    enabled: false
```
Ameliorate, line 205: This disables the module that disables phantoms. You want to leave this as `enabled: true`.

Ameliorate (changes requested):

templates/public/plugins/SimpleAdminHacks/config.yml.j2
Protonull: That's not true, this would disable the hack that *regulates* phantoms.. doing this would actually return phantoms to their vanilla behaviour.

TheOrangeWizard: does PhantomMenace enabled: true and night+storm spawn:false do what i want it to do

Ameliorate: Indeed. However, the goal of this PR is to disable phantom spawns completely.

Protonull: @TheOrangeWizard Yes, that would disable phantom spawning.

Ameliorate:
I agree with this change. No use has been added for the drops of phantoms. In effect, phantoms exist purely to wear down your prot dura while you're traveling. There is no way to prevent phantoms from spawning, besides doing an annoying action that sets your spawn to wherever you are at the current. This PR might be less valid if someone made a patch that disabled phantom spawns in well-lit areas or within bastions, but there is no such change incoming. Even with the current vast nerfs to phantom spawning, phantoms still feel like they spawn way too much for an annoyance """feature""".

This message is a personal attack on Orinnari specifically, despite what the content may suggest. Orinnari can and should take offence to what I am saying.

Protonull:
Something that I added [here](https://github.com/CivClassic/SimpleAdminHacks/blob/417459afaf7453c0c468e1a9a03e9e9157cb638d/src/main/java/com/programmerdan/minecraft/simpleadminhacks/hacks/basic/PhantomMenace.java#L78-L93) is a light level limiter. Phantom spawn behaviour is a little unorthodox since it doesn't *seemingly* interact with light levels the way other mobs do\* and instead relies on a player's insomnia level and block placement. This change regulates phantom spawning to bring it more inline with other mobs, so if you're standing in a lit area, phantoms shouldn't spawn on you. [This change was merged yesterday](https://github.com/CivClassic/AnsibleSetup/pull/114/files), so I'm unsure whether it's on the server yet, but I am listening to people's feedback to balance phantoms without flatly removing them.

templates/public/plugins/SimpleAdminHacks/config.yml.j2
Protonull: > Players have a statistic called TIME_SINCE_REST. If a player's value is above 72k (3 days minecraft time) then they will start spawning phantoms. **Setting the cap to below 72k will effectively disable phantoms.** Setting the cap to -1 will disable the cap. Default: 80k (10% chance per two minutes during the night and thunder storms.)
> Read more here: https://minecraft.gamepedia.com/Phantom#Java_Edition

You could also do:
`timeSinceRestCap: 0`

if you really want to redundantly disable phantoms

TheOrangeWizard:
lighting - one would have to light up the entire surface area within mob ai range, while skeletons etc. can be foiled by a simple wall phantoms will fly right over it, so phantom-proofing your farm/whatever takes quite a bit more effort than regular mob-proofing.

bastions - preventing phantoms from spawning in bastions alone has the same problem as lighting, ideally i think phantoms would die/cull upon entering a bastion field if we were to take this approach.

this is also assuming conditional phantom spawning is desirable of which i remain unconvinced

Protonull:
Something that I have been mulling over is whether to have beacons that emit regeneration hurt phantoms, and perhaps even skeletons and such. It would mesh quite well with the established mechanic of healing doing damage to undead mobs, and would incentive players to build more infrastructure in more permanent settlements.

Maxopoly:
Rejecting, completely disabling phantoms is not desired atm

Protonull:
based

TheOrangeWizard:
tragic

PR #120, opened 2020-12-04
add sticks -> chest carpentry recipe

TheOrangeWizard:
leaves now drop sticks which are not consumed in any great quantities unlike logs+leaves, meaning tree farming now produces many relatively useless sticks as a byproduct

allowing sticks -> chests effectively provides a minor buff to tree farm wood output

this recipe is 1:1 with crafting chests from planks which could be changed if necessary

### DIFF
--- a/templates/public/plugins/FactoryMod/config.yml.j2
+++ b/templates/public/plugins/FactoryMod/config.yml.j2
@@ -430,6 +430,7 @@ factories:
      - make_beehive
      - make_casing
      - make_crates
+     - sticks_to_chest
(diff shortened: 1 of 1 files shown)


Protonull:
Why not just a sticks -> planks recipe?

TheOrangeWizard:
because chests are the primary ingredient in carpentry recipes, making planks adds an extra crafting step

also seems to make sense to keep them slightly more limited in application

Maxopoly:
Seems alright

PR #121, opened 2020-12-04
add flint -> sand grinding mill recipe

TheOrangeWizard:
flint is an often unintentionally collected byproduct of gravel, with little use

dunno if getting sand from flint makes any sense irl but it matches the gravel -> sand grinding recipe, including the 1:1 ratio which could be adjusted

### DIFF
--- a/templates/public/plugins/FactoryMod/config.yml.j2
+++ b/templates/public/plugins/FactoryMod/config.yml.j2
@@ -720,6 +720,7 @@ factories:
      - crush_stone
      - crush_cobble
      - crush_gravel
+     - grind_flint
      - grind_sandstone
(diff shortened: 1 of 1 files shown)


Protonull:
Definitely agree with a flint sink, but the 1:1 ratio you propose is definitely far too high

wingzero54:
I understand wanting to find a use for flint, but it just doesn't make sense for flint to be ground into sand and we don't need to add any more buffs to sand production

Protonull:
Sand production through factories should be heavily nerfed, because remember that stone is a renewable resource and so is the charcoal. Sand is quite literally now an infinite resource to anyone with a rudimentary bot.

Issue #51, opened 2020-07-14
obfuscate to a non-stone block depending on surrounding blocks


Gjum:
this was brought up by max in #50:

> per biome overrides for the nether? With this setup any snitches etc. would still stick out in all the netherrack

the same goes for building blocks (obsidian) and different y levels (sand, dirt, ...), so it's not really biome specific, but rather dependent on surrounding blocks.

possible methods I brought up in the past were:

- pick the most common block type out of the surrounding 6 face-adjacent blocks (or 5 (4) if at a chunk border (corner))
- always pick the block type north of the obfuscated block (north is arbitrary here; and in case of chunk boundary, pick any other block face in precedence order)

in case the adjacent blocks are also obfuscated blocks, treat them as out-of-world like above, or fall back to stone, etc.

does orebfuscator have an option for this? if not it might be worth it moving this issue to its repo.

what's the performance hit? if it's non-negligible, is it worth it vs. just always obfuscating to stone/biome-default?

Nuclear-Crow:
If you still want to fall back to default when something like [this](https://prnt.sc/tib2h2) comes up, why bother with adaptiveness in the first place

Gjum:
in your screenshot, each obfuscated block still has an adjacent stone block, so with both methods they would all be obfuscated correctly; maybe my explanation was unclear. in any case, this is a heuristic, so breaking in extreme cases (e.g., 3x3 cube of noteblocks) isn't a problem with respect to the overall improvement.

do you know other edge cases? or other methods of block type selection that catch more cases or are easier on performance?

Maxopoly:
Implemented by https://github.com/CivClassic/AnsibleSetup/pull/123

PR #123, opened 2020-12-10
cleaned up and adjusted in line with old config

Quinndolin:
removed orbfuscation sections for nether and end - not present/enabled in old config
enabled fastgaze and set proximity distance to 8
readded jukes noteblocks and sponges to proximity, also added bone blocks
proximity uses block below by default, this seems desirable so the end and overworld can share a config
added air to orebfuscation

### DIFF
--- a/templates/public/plugins/Orebfuscator/config.yml.j2
+++ b/templates/public/plugins/Orebfuscator/config.yml.j2
@@ -65,91 +65,15 @@ world:
       obsidian: 1
       redstone_ore: 1
       stone: 1
-  - worlds:
-      - world_nether
-    enabled: true
-    darknessBlocks:
-      - chest
-      - ender_chest
-      - trapped_chest
-      - spawner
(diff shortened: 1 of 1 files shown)


Maxopoly:
>  proximity uses block below by default, this seems desirable so the end and overworld can share a config

So you can just replace the block below you, relog and check which blocks changed to get a list of all "interesting" blocks within view distance?

Quinndolin:
block below the block that is being obfuscated not the block below the player if im understanding what you're saying

Quinndolin:
its a shame that a 'most common adjacent block' setting isn't in there then things could obfuscate to obfuscate to obsidian in vaults

Gjum:
> 'most common adjacent block' setting

tracked here https://github.com/CivClassic/AnsibleSetup/issues/51

Maxopoly:
Ok cool

PR #126, opened 2020-12-15
Remove TradePlus

Protonull:
It was a mistake to add this plugin, sorry about that D: And if you see the state of its source code, it'd need a full rewrite to really make it viable. For the time being it should probably be removed.


Maxopoly:
Ok, but what's wrong with having it in it's current state?

Protonull:
TradePlus *should* offer a way to allow you to trust what you're trading in that you can see, in the GUI, what exactly is being traded for what and in what quantities. This differs from the usual method of just throwing what you both have at the ground or at each other.

However, TradePlus seems very buggy. For example, you can't trade compacted items despite not being restricted in the config, which by itself makes it useless for bulk trades, arguably the biggest use case for the plugin. And I would not be surprised if there was an item destroyer / dupe bug in there, especially given this commit https://github.com/Trophonix/TradePlus/commit/851135e952ca16565a1a33a6bc83d29455bf7b86 which we haven't merged yet.

Maxopoly:
Hmm alright

PR #136, opened 2021-01-03
Recipe Request, Split Snow Blocks into Snow "Slices"

KaloHG:


wingzero54:
You can craft snow blocks into snow in crafting tables vanilla, is it necessary to add a factory recipe for it?

KaloHG:
Oh, I was completely unaware of this recipe. It should be noted this would be much more effecient than that recipe though. Instead a 1:1 instead of 1:0.5

PR #138, opened 2021-01-03
Update config.yml.j2

Quinndolin:
removes shulker boxes from proximity hiding (non-existant, even if they were they spend most of their time in inventories anyway)

removes bone_block from proximity hiding
removes diamond and emerald ore from proximity hiding
(pointless to have these, only ruins people's builds)

### DIFF
--- a/templates/public/plugins/Orebfuscator/config.yml.j2
+++ b/templates/public/plugins/Orebfuscator/config.yml.j2
@@ -75,8 +75,6 @@ proximity:
     distance: 8
     useFastGazeCheck: true
     hiddenBlocks:
-      diamond_ore: {}
-      emerald_ore: {}
       chest: {}
       ender_chest: {}
       trapped_chest: {}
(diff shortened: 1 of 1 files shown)


Ameliorate:
Emerald ore is kind of iffy because it spawned in extreme hils biomes at SOTW. If this were merged I could go into an extreme hills biome and xray for emerald ore.

wingzero54:
I'm okay with removing Bone Blocks as there are a lot of builds using bone blocks and city bastions seem much less likely to be xrayed than vault bastions. As for shulkers yeah I agree that people likely wouldn't just be keeping them out and I'm not sure they need to be obscured.

Maxopoly:
>  removes diamond and emerald ore from proximity hiding

Not merging as long as that is in there

Quinndolin:
emerald ore would still be obfuscated just not proximity hidden

afaik this shoudln't make mining for emeralds in extreme hills with xray any easier, you will still be greeted with a wall of ores when you turn x-ray on, whether the emerald is getting proximity hidden to stone behind that wall or not shouldn't matter

running through caves with x-ray might let you clear caves quicker thats all i can think of

wingzero54:
If you strip this down to just bone blocks and shulkers I'll accept

wingzero54:
Unnecessary with [this commit](https://github.com/CivClassic/AnsibleSetup/commit/96fcc30d41f386b0f785ba6602e35f2537886a08)

PR #145, opened 2021-01-27
Remove Diamond Veins, The Failed Experiment That Kills Newfriends

Ameliorate:
This change removes diamond veins and replaces them with a flat chance to find 1-3 diamond ore every 1842 stone blocks (on average, rates copied from devoted3 hiddenore default config).

The diff is kind of messed up because git. You should click [view file](https://github.com/Ameliorate/AnsibleSetup/blob/0e3af7f94dd2bb33cfdd7a1a7c00df6f11abb451/templates/public/plugins/HiddenOre/config.yml.j2#L148-L158) to see the actual changes.

---

# Why?

## They cause newfriends to quit.

Diamonds seperate newfriends from the real players. If you have diamonds, you are able to play the game, purchase goods, mine with much higher efficency, and actually build. Many times I have seen newfriends quit over being unable to obtain diamonds. In Icenia, the goal we give all newfriends is to understand mining and obtain a silk touch pick. Those newfriends that try to learn to mine, following exactly one piece of good information on how to mine on the subreddit (which by the way is flawed), complain endlessly that diamonds are too rare, and that you need to mine for far too great a distance in order to find a vein. I agree with this thought based on my experince mining.

Now at this point people will argue "you just have bad luck". I argue that this is the case. I argue that this is also a bad thing, a very bad thing.

Suppose I have bad luck whenever I mine. Suppose I'm a newfriend. I will not wait out my bad luck. I will go on a server where mining is actually fun, where bad luck averages out far quicker because it doesn't have a failed experiment from a six year old server that was emulated out of a failed goal nearly four years ago.

## The Failed Experiment: Diamond Veins Don't Encorage Trade

Veins were added to CivCraft 2.0. The idea was that everything in the game was a vein. This actually worked. It worked good. Because iron was a vein, and diamond was a vein, and coal was a vein, it meant that you would trade all of these resources between each other, and they all had value. If you wanted iron, you downloaded a """minimap""" and went to ice mountains. If you wanted diamonds, you would enable """waypoints""" and """toggle a waypoint""" whenever you found an ore. It almost worked well.

**XRAY**

Oh yeah, there was that little fact. Everyone on civ2.0 xrayed. Everyone. Xray was, in effect, a required mechanic for mining.

Seeing this, when TealNerd was first creating the configs for CivClassics, he likely wanted to try it again and redo the experiment of veins. However when he was doing this, teal made a mistake. I think that the civ2.0 system where everything was a vein was actually a good idea. However, teal's system where the only vein type is diamond is flawed. You see, on civ2.0 you would go and mine iron. You would go mine coal. You would go mine diamond. You then would exchange your iron for diamonds and coal, or exchange your coal for iron and diamonds, or your diamonds for iron and coal. On civclassics nowadays, you go and mine diamonds. As a result of mining for diamonds, you randomly dig up iron and coal. You now have diamonds, iron, and coal. You don't do any trade. 

## Diamond Veins Have Lost Our Trust

Diamond veins have been changed a lot. A lot alot. And about half the time when they're changed, they're actually broken. Across updates, changes, and """upgrades""", about half of them have broken diamond veins for months at a time. To make matters worse, diamond veins are incredibly hard to debug. As a result, diamond veins, when broken, remain broken. It is incredibly hard to edit and reason about diamond veins in the config.  Throughout civclassics history, there have been diamond veins have been widely known by the community to be broken for months at a time.

## Diamond Veins Aren't Fun

When you're designing a game, you should ask yourself three questions:

* Is it balanced in consiteration with all the other mechanics in the game?
* Do the players feel like it is balanced?
* Do the players feel like it is fun?

Diamond veins are balanced, consitering all the other mechanics in the game.

The players are in agreement that diamond veins are balanced.

Are diamond veins fun though?

I argue they aren't. I argue that replacing them with a flat chance to drop a diamond ore every time you break a stone block is more fun. When I used to go mining on devoted 2.0, I could watch anime on my second monitor and mine diamonds all day long. However on civclassics, even if I can watch anime while searching for a vein, I can't while mining out a vein. So now we need to ask ourselvess if mining out a vein is fun, because I can't watch a fun anime while doing so. 

There are three things that make mining out a vein not fun. All of these stem from a lack of information and too much concentration. You don't know where the diamond ores were unless you've placed markers when they spawned. You don't know where to dig next, and where the vein is going, although you can guess based on your markers. Finally, you don't know if you're done diging out a vein. In civ2.0 none of these were an issue, because the vein was blocks that have been placed in the world. Sure, you can't see where the ore blocks were, however you can see where the next ore was (even if you don't xray), you can see where the next ore blocks are, and you know when you're done when you don't see any more ore blocks on xray. In addition, you could strip mine out veins on civ2.0 like vanilla if you don't use xray. You can not strip mine on civclassics.

# What is it being replaced with?

It's being replaced with a very simple mechanic. Just like coal, and iron, and gold, and redstone, and lapis, whenever you break a stone block below y=16, the game rolls a dice with about 1800 sides. If it lands on a 1, the game will spawn in between 1 and 3 diamond ore blocks near the block you've just broken. In this way, the most efficent way to mine diamonds, as well as the rest of the ores is to dig a long tunnel, or a big hole. In this way, newfriends mining the same way they're used to on vanilla will find diamonds, without having to be taught the great art of *diamond mining*.

# TLDR

Diamond veins have been on civclassics for nearly four years. In those four years they have been found to have some problems. This PR, if accepted, will remove veins and replace them with the mechanics found in Devoted 3.0 and Civ3.0.


TheOrangeWizard:
Blessed

ProgrammerDan:
the big success of civ2 veins wasn't prevalence of x-ray, as orebfuscator existed then (although had some bugs, and a large deobfs radius that made x-ray mining _advantageous_ but not instantaneous) but that the diamond rate never really went to 0, and diamond veins were straight common. I personally found several and they were individually massive; worked with others to mine even more and each was, again, massive. Let's just say that people who were diamond poor in Civ2 were diamond poor because they never mined, not because diamond was rare.

There's no reason that hiddenore can't be configured to be equivalently productive. You can even use BOTH configs -- a low level "always around" diamond ore chance, with concentrated ore availability in diamond veins.

The big issue is definitely if they are too concentrated in veins, and too rare outside of veins, and veins themselves are uncommon -- this is shitty, and needs to be fixed. Gotta keep that dopamine drip going, or you're right -- players will leave. And they should, that's crap. Keep that drip, keep them hooked, keep them mining.

Ameliorate:
Another interesting point Naglifer brought up to me last night was that powerplayers like himself and NeoTide don't trust diamond veins anymore, and because of that they don't spend diamonds unless they have to. Even if diamond veins may be working now, they can't trust that they'll be working in the future. As a result of that fact, even if they want to, they will not spend diamond on luxuary items like map art and building blocks. To these players, diamonds are a vehicle to get more diamonds, by selling exp, prot, and enchanted tools. As a result, if they spend too many diamonds and run out, and if veins are broken at the time, they will have no way to get those diamonds back. 

In this way, veins actually slow down the economy, because the powerplayers with large amounts of liquid wealth won't spend their diamonds on luxuary goods.

Ameliorate:
@ProgrammerDan 
\>There's no reason that hiddenore can't be configured to be equivalently productive. You can even use BOTH configs -- a low level "always around" diamond ore chance, with concentrated ore availability in diamond veins.

I think this is a bad idea. If I dig up a diamond ore while mining, how do I know if it is a vein or just a random sparse diamond ore? Sure, you could have one chat message when it's a sparse diamond and another when it's a vein diamond, but I don't want a feature whose functionality depends on chat messages I randomly get while mining.

\>Gotta keep that dopamine drip going

I agree. However, I think that large, sparse veins are a bad way to go about keeping the dopamine drip going. They are hard to teach to newfriends, and have all the flaws I've argued for above.

ProgrammerDan:
@Ameliorate 

>both

CivCraft 2.0 was set up as I described, which you referenced as a successful implementation. They didn't use HiddenOre, but their exact model can be replicated using HiddenOre. That's the only reason I mention it :)

> dopamine

Yeah, I don't really have an opinion. I'm aware of how Tealnerd designed his veins, and my updates just made sure they actually worked the way he imagined they needed to work. From a configuration standpoint I'm a big fan of constant dopamine drip, which is why Devoted never used veins, we only used percent chance with clusters. Much better mental payout.

Veins can be done right but it's much, much harder, and should be a supplement to a constant drip (BIG WIN style) never a replacement. We are in agreement.

Maxopoly:
I don't feel strongly one way or the other about this

wingzero54:
I would rather go back to old veins than regular hidden ore spawning. However I understand that it is confusing to new players who have to learn two totally different styles of mining, and it would make sense to make it all one style.

ProgrammerDan:
I recommend a low yield general diamond cluster spawn, with high density
veins, emulates civ 2 and is best of both. Strip mining and lucky strikes
both become valuable, everyone wins across all styles. Basically just add
what was suggested here instead of replacing veins. Rebalance probabilities
to fit intended diamond available balance, if needed. Win, win.

AllenY:
aye

wingzero54:
It seems to me pretty much all complaints went away when I put diamond veins back to how they used to be. If this is still desired, we have an opportunity to use both veins & random drops using deepslate diamond (when we eventually upgrade to 1.17) to differentiate the two to avoid confusion. A new PR should be raised then. I do not support removing veins entirely

PR #148, opened 2021-02-16
Add impossible banner pattern recipes

Lodish:
Two banner patterns are impossible to obtain on classics, the globe pattern that requires trading with a villager and the piglin pattern that can only be found while looting [Bastion Remnants](https://minecraft.gamepedia.com/Bastion_Remnant). This adds recipes to craft them in the Printing Press.

### DIFF
--- a/templates/public/plugins/FactoryMod/config.yml.j2
+++ b/templates/public/plugins/FactoryMod/config.yml.j2
@@ -661,6 +661,7 @@ factories:
       - repair_printer
       - craft_globe_banner_pattern
       - craft_piglin_banner_pattern
+      - craft_creeper_banner_pattern
   concrete_mixer:
     type: FCC
     name: Concrete Mixer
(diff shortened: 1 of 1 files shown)


wingzero54:
What about using a piglin spawn egg instead of blackstone? Also a creeper head banner recipe with creeper egg?

Lodish:
I guess you could, it was just that the Chiselled Blackstone block looks pretty close to the pattern on the banner pattern.
Chiseled_Polished_Blackstone_JE1_BE1 It has that snout sort of look.

I didn't add the creeper banner as I think it can be created on the server already.

wingzero54:
Fair enough it is very close to the pattern. And creeper banner may hypothetically be possible, but I think it is worthwhile to add a recipe for it

wingzero54:
So villager eggs are quite rare which make the globe pattern quite rare, whereas the snout pattern is much easier. We should establish whether all the banner patterns should be equally rare and expensive still, or equally cheap and accessible

Lodish:
Banners patterns are infinite use, and you only need to make a banner once, you can copy that banner forever using the base version of a banner. So someone could rent an egg or pattern, or someone with a villager egg could makes lots of banners patterns and sell them (the recipe returns the villager egg in the same way as trading with a villager doesn't make the villager disappear). I don't know if you can anthropod farm villagers for eggs but there's probably a few collecting dust in chests via fossils. Banner patterns are pretty varied in their rarity, for instance the flower banner pattern just requires finding an oxeye daisy, the mojang one you need to search on average 71 mineshaft chests to get an enchanted golden apple (assuming this number hasn't changed since 1.10) the mojang one is easier now as you know need one apple for infinite banners, I know people that used up enchanted apples to do one banner early on in the server. 

I'm happy to make the piglin one more expensive though. I suppose in vanilla searching bastions involves fighting piglin brutes, don't know how to incorporate that into a recipe though.

Lodish:
Also in regards to also adding a creeper banner pattern recipe, I'm curious if AllenY will able to get a legitimate one
image

wingzero54:
I'm thinking make the snout a whole stack of blocks and add a creeper banner pattern and I'll merge it.

Lodish:
It's now a whole stack of blocks and I've added a creeper banner pattern.

PR #155, opened 2021-03-12
Tweak carpentry factory rates, add carpentry factories for all wood types

squareblob:
This PR is comprised of two changes

### Tweak Carpentry Factory Rates (6cccb0b)

Carpentry factory rates are still roughly three times more efficient than vanilla crafting but the variance is now much less (see image below). A recipe for fence gates was added and the recipes are adjusted to produce 128 output.

Carpentry Factory Rates

### Add Carpentry Factories for all Wood Types (ddb5d03)

Now onto the considerably more janky change.

I have kept the current carpentry factory producing oak output and added seven new carpentry factory variants for all other wood types. These variant carpentry factories produce fences, signs, trapdoors, doors and fence gates using logs of said type. They take four stacks of variant wood to setup (versus 32 chests for regular carpentry) and one stack of variant wood to repair (versus 2 chests for regular carpentry).  

Some alternatives to this change are :

1) Put all variant wood type recipes in a single factory (such as the current carpentry factory). 

This does not solve much of the jankiness because you still need seven duplicate recipes for each wood output, making modifying recipe rates a bit more annoying (among other things). This method would also remove any possible specialization in wood types with no wood variant specific repair or setup costs. Nevertheless, I'm rather neutral between this and the current PR, and would be happy to do this if desired.

2) Codeityourself a way to have a single Factorymod recipe work for multiple wood variants

eg. It would be neat to specify [tags](https://minecraft.gamepedia.com/Tag) in recipes. You could use built in tags like `logs`, `wooden_doors`, `wooden_fences` etc. But having a flexible system like this would require a lot of changes (will not happen) and would probably introduce its own jankiness anyway.


wingzero54:
oooh graphs

wingzero54:
I know it's desirable to have all the wood variants available through factory, but I hate that it has to be done by creating 7 duplicates of the carpentry factory. One factory with 48 recipes wouldn't be too bad, would it?

PR #156, opened 2021-03-14
Duplicate others: biomes

patrick1bauer:


okx-code:
This would break the config for crops that do not define a growth speed for `end` or `nether`. Usages of `other` in the crop configuration should be replaced with `end` and `nether`.

PR #157, opened 2021-03-15
Fix DogFact

Diet-Cola:
We cant use apostrophe's without escape characters since we use them to encase the string


Diet-Cola:
Just tested, cant even use escape characters on apostrophes 🤔

caucow:
check docs
> When using single-quoted strings, any single quote ' inside its contents must be doubled to escape it:
> `'A single quote '' inside a single-quoted string'`
[see here](https://symfony.com/doc/current/components/yaml/yaml_format.html#strings)

PR #158, opened 2021-03-15
Create github action to lint YAML

### LINKED ISSUE #108: Setup Github actions to validate YAML files
Might be a bit complicated, because their file ending is .yml.j2

AngrySoundTech:
I left almost all the rules for the lint default aside from making line length and trailing spaces warnings instead of errors, but the rules are up to you obviously.

Because this PR is only creating the workflow, the workflow will not pass initially because of pre-existing validation errors that should be fixed after merging.

Rules for the config_data block can be found [here](https://yamllint.readthedocs.io/en/stable/rules.html), let me know if you want any of them changed for this PR.

Closes #108

### DIFF
--- a/.github/workflows/lint-yaml.yml
+++ b/.github/workflows/lint-yaml.yml
@@ -1,0 +1,24 @@
+name: Lint YAML
+
+on: [push]
+
+jobs:
+  lint-yaml:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


wingzero54:
I need somebody smarter than me to do this, would be a good config checker to ensure we don't break plugins by having tabs instead of spaces, or irregular formatting, etc

PR #159, opened 2021-03-17
Removed "other" biomes

patrick1bauer:
Removed duplicate "other" biome type from the biome list and changed crops that grew in the "other" biome to grow with the same rate in the "nether" and "end" biomes.


patrick1bauer:
Just cleaning up the config a little

PR #168, opened 2021-04-07
Enable wither spawning

Diet-Cola:
One to test but I believe there is nothing game breaking having withers enabled, might be one to test

### DIFF
--- a/templates/public/plugins/SimpleAdminHacks/config.yml.j2
+++ b/templates/public/plugins/SimpleAdminHacks/config.yml.j2
@@ -256,7 +256,7 @@ hacks:
     enabled: true
     disablePotatoXP: false
     villagerTrading: false
-    witherSpawning: false
+    witherSpawning: true
     patrolSpawning: false
     phantomSpawning: true
     enderChestPlacement: true


KaloHG:
considering how wither skulls are relatively common & farmable to some extent, my main concern is people unleashing like 20 withers on a skybridge if a vault were ever skybridged lol

Diet-Cola:
Ends up as a 50/50, the withers could absolutely turn around and focus you

wingzero54:
We tested some and Withers seems to obey citadel, the only thing is they are pretty obnoxious and OP. But I see that as an absolute win :glad: only reason to keep them disabled (unless some random issue comes up) is them just being obnoxious to deal with. They're like carpet bombers

wingzero54:
> 
> 
> pls city bastion prevent wither spawning if not on group if this gets enabled

Hm yeah that makes a lot of sense. Would need a dev to look into if you can cancel wither spawn even based on player that started the event and their bastion check

caucow:
> > pls city bastion prevent wither spawning if not on group if this gets enabled
> 
> Hm yeah that makes a lot of sense. Would need a dev to look into if you can cancel wither spawn even based on player that started the event and their bastion check

Don't city bastions prevent placing the blocks required to spawn a wither? Or are CC and CR city bastions different somehow? If the bastion prevents block placement to people that don't have build perms it'll prevent player-placed wither spawning. Only other way to spawn them afaik is with dispensers, which *still requires players to place the soul sand*, which would generally be more useful as a trap than pure offensive destruction.

wingzero54:
City bastions only prevent reinforced block placement, you can walk into a city bastion and place whatever unreinforced blocks you want

caucow:
Ah. Well in that case, you'd have to check the precursor BlockPlaceEvent to get player info, and probably outright disable BlockDispenseEvent if the block is a dispenser and the item is a wither skull since redstone triggered stuff can't be linked to a player. Both are precursor events to CreatureSpawnEvent which only knows the entity being spawned and the SpawnReason, so you'd lose the ability to get any player info if that's the only event you covered.

Diet-Cola:
Reviving this, maybe being lazy but I dont think its worth the effort of tracking what the player is placing & preventing dispenser actions to disable wither spawning in city bastions. We can easily prevent anyone building a wither in a city bastion regardless of group perms as a good compromise.

Convoy20:
>One to test but I believe there is nothing game breaking having withers enabled, might be one to test

Yeah this doesn't sound like good idea

Diet-Cola:
Works from dev hell & limited sessions on civtest

Convoy20:
> Works from dev hell & limited sessions on civtest

People don't have stacks on stacks of wither heads on dev hell... I don't see the benefit of adding this

Diet-Cola:
We can have phantoms but not have withers?

Convoy20:
no comment

PR #177, opened 2021-04-16
Delete Old Banstick

Diet-Cola:


Diet-Cola:
There are two bansticks in the plugins directory

Protonull:
And two KiraBukkitRelays

Diet-Cola:
Ahh so do we delete 1.2.1 which was uploaded 4 months ago or 1.2.0 which was 21 days ago 🤔

wingzero54:
The newer one is correct, it has the cmc update. The older one is just from the 1.16 update, and Max marked the version as 1.2.0 so not sure why the jar was 1.2.1

PR #182, opened 2021-04-19
Buff slimes

KaloHG:
currently slimes rarely spawn, and will immediately despawn if they've been alive longer than 1200 seconds i believe? which is pretty atrocious when attempting to farm them whatsoever.


KaloHG:
*1200 ticks (60 seconds)

PR #185, opened 2021-04-29
Mildly Reduce Mob Despawning

KaloHG:
32 blocks for passive despawning is far too low :/
We double it, since it will only increase the total amount of mobs spawned by ~25% (mobs currently are no issue in regards to server memory or processing.)
However just in-case we'll reduce hard-despawn to 100.


wingzero54:
I want to continue making improvements to the server before easing up on mob spawns. At 60 players the TPS starts dropping, if we ease up on mobs it'll only make it worse

PR #186, opened 2021-05-02
Add compact repair kits

Ameliorate:
You can prepay the materials for a compactor repair to make a compact repair kit. You then can use a compact repair kit to repair a compactor.

This resolves a common complaint about compactors that it is annoying to constantly repair them.

### DIFF
--- a/templates/public/plugins/FactoryMod/config.yml.j2
+++ b/templates/public/plugins/FactoryMod/config.yml.j2
@@ -685,7 +685,9 @@ factories:
     recipes:
       - compact
       - decompact
+      - make_compact_repair_kit
       - repair_compactor
+      - repair_compactor_kit
   printing_press:
     type: FCC
     name: Printing Press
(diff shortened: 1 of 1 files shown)


wingzero54:
So the whole point of this, is instead of leaving redstone + piston + iron block in your factory, you just leave the repair kit instead? I can't say I understand the value of this, I just always left the repair mats in the factory. This is leaving the repair mats, just with an extra step

Ameliorate:
The reasoning behind this was for repairing compactor factories that you're traveling to, that you know are in disrepair and don't have any repair materials in them.

Protonull:
As long as the cost of a single repair kit is the same amount as the cost to actually repair the factory, this is a great idea. That being said, I think this could also be solved through markets. If ItemExchange for example allowed for multiple items, it would mean players would be able to buy a "Repair Compactor" set in one trade.

okx-code:
Please add this, it would make my life so much easier

wingzero54:
I'll allow it but I'm curious how much it gets used long-term

PR #188, opened 2021-05-03
Remove villagers after they've been alive for 30minutes

Convoy20:
Remove villagers after they've been alive for 30minutes


wingzero54:
I just don't see this as necessary

PR #189, opened 2021-05-03
Alt-item despawn changes

Convoy20:
Adding Netherite update items so they don't despawn quickly.

Drastically lowered time to cause cobblestone to despawn (600 = 30 seconds, was previously at 6000, or 5m... which is default)
Added in other common drops that don't get picked up while mining dirt or stone.


wingzero54:
I'll add extended netherite decay but not really interesting in having such fast despawn for the other random items

PR #190, opened 2021-05-04
Change Exilepearl to tick less regularly. Sue me

Convoy20:
Exilepearl ticks every 8hrs

### DIFF
--- a/templates/public/plugins/ExilePearl/config.yml.j2
+++ b/templates/public/plugins/ExilePearl/config.yml.j2
@@ -19,8 +19,8 @@ pearls:
   hotbar_needed: true        
   decay_interval_human: week                                                   
   decay_interval_min_human: 10080                                              
-  decay_interval_min: 15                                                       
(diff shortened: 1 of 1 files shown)


wingzero54:
@Diet-Cola really, is there any downside to just ticking pearls less like this? Pearl health will change a little more drastically, but it will still be updated 4 times a day

Diet-Cola:
I remember looking at EP stuff and it checks if a pearls health is equal to 0, not less than or equal so we have to make sure it will properly handle a potential sub-zero health from stuff like this

wingzero54:
Optimal solution would still be to spread the pearl ticking task out better

Convoy20:
> Optimal solution would still be to spread the pearl ticking task out better

Why not merge this for now until that task gets assigned and completed?

wingzero54:
Not the way I wanna go about it, there's a PR for adjusting pearls so the tick won't be intensive

PR #192, opened 2021-05-04
ClearLagg entity removal turned on.

Convoy20:
More of a sanity check, but will activate very infrequently and remove any projectiles it can find.


wingzero54:
I won't run these "delete all of X", even if they are on a long timer. Not good practice and enables bad habits

PR #193, opened 2021-05-05
RIP everyone seeing advancements

Diet-Cola:


wingzero54:
Is this really a problem though? We're a multiplayer server, and it was pretty amusing for everybody to see the achiev of a player 4 years in

PR #194, opened 2021-05-07
Allow "illegal access"

Protonull:
Prevents warnings and errors from ProtocolLib for perfectly ordinary uses of reflection that Java is attempting to clamp down on.


wingzero54:
Not updating to java 16

Protonull:
Not right now but maybe at some point in the future. And Java 11 *does* spit out unnecessary warnings about reflection that this flag would fix.

PR #195, opened 2021-05-13
Change "chests" to "crates" in compactor

MrJeremyFisher:
I am unable to test this at the moment, but judging by how this is used in other places it should work. Although I would recommend testing it on the test server or something


Ameliorate:
This will not work. Factories require every NBT tag in the item to be exactly the same as what they are looking for. If the item is renamed it will not accept the item.

MrJeremyFisher:
Ah I see, now I can actually test this aswell, thanks for the tip.

PR #196, opened 2021-05-18
Make uncarving pumpkins consume no fuel

Ameliorate:
According to a conversation I had in globalchat, the recipe is currently consuming fuel, at a rate of 1 charcoal per pumpkin. This seems to have been an oversight.

This change makes uncarving pumpkins take no fuel.

### DIFF
--- a/templates/public/plugins/FactoryMod/config.yml.j2
+++ b/templates/public/plugins/FactoryMod/config.yml.j2
@@ -4376,6 +4376,7 @@ recipes:
   uncarve_pumpkins:
     forceInclude: true
     production_time: 1s
+    fuel_consumption_intervall: 99999s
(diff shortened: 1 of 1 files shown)


Protonull:
But why is the solution for it being too expensive making it completely free?

Ameliorate:
Because this recipe simply for an oversight that before 1.14 you got carved pumpkings while farming, while now you get uncarved pumpkins.

Diet-Cola:
How about we remove this recipe and make it a crafting recipe using CraftEnhance, I personally hate small recipe things like this in a factory

Ameliorate:
bd24ac3dd988dc07db1dcb5778dd712496d2c64c is an adequete solution, so I'll close.

PR #199, opened 2021-05-26
Proposed pearl cost change

wingzero54:


Convoy20:
Why make players who were pearled for hacking/vpning/jumping into vault holes suddenly start costing ridiculously more.

Why wouldn't a player who gets imprisoned on an alt immediately go radio silent and just wait for the costs to force their captors to free them.

This doesn't allow you to apply justice, this adds costs, forces the unjust to go free, and overall is just fucking dumb.

We had expensive pearl costs on CivRealms. How did that go?

CivUniverse didn't keep anyone pearled for longer than a week. How did that go?

Do you have any fucking data showing that this is a good idea or are you just getting swayed by people who have been pearled for years and years who turn around and do the same exact bullshit they got pearled for on server after server?

There are a handful of pearls that are at best *contentious* at this point. Icenia's vault audit presentation showed that most of their pearls that get held are simply ones that have never rejoined or never communicated with anyone. If they don't make the effort to get freed, they don't deserve to be freed.

A vast amount of pearls are players who raid, attack innocent towns, grief, obbybomb, etc. Why does it makes any sense to make these pearls start costing more, especially when those who are held for a long period of time are being held because they deserve, absolutely deserve, to be removed from the community. 

Why are you putting more work on the shoulders of your player base, that is attempting to ensure that people can build and play in peace, to make sure your server doesn't get burnt to the fucking ground.

If you want to kill off CivClassics because it's becoming stressful, just quit the admin team.

wingzero54:
In your opinion, how many pearls should one player be able to hold by themself?

Convoy20:
In my opinion. Unlimited. It's the community that decides how many pearls a player should hold. Not admins. It is not a balancing call, it is the literal name of the game, civilization.

Protonull:
> In my opinion. Unlimited. It's the community that decides how many pearls a player should hold. Not admins. It is not a balancing call, it is the literal name of the game, civilization.

That would seem, at least in my humble opinion, highly optimistic to put it politely. Would you actually and in good faith say that the "community" (which I assume to mean the player-base at large rather than the inner circles of alliances) has an undeniable and informed role in the accountability of pearls? Perhaps now with the introduction of `/ep showall` this is more the case, but would you say that the current setup yields regularly and effectively as to genuinely posit that the community does indeed "decide how many pearls a player should hold?"

It's all very well and good saying that the players should decide among themselves how many pearls should be held, and I would generally agree with that statement, but if that's not happening then doesn't that at the very least imply an imbalance between the premise of the pearling and the realities of pearling?

Convoy20:
In case you haven't noticed, there have been about 50 wars waged on various civ servers over the course of its history to determine that very question.

It happens. It has happening. If you have an issue with it on classics, start looking at the awful vault meta, not arbitrarily freeing pearls through balance changes @Protonull

Protonull:
> It happens. It has happening. If you have an issue with it on classics, start looking at the awful vault meta, not arbitrarily freeing pearls through balance changes.

Oh I have not discussed any of this with Wingzero, nor even investigated the change. I was specifically replying to your opinion that the admins should intervene to allow infinite pearls on the premise that there's some established community based form of accountability. And you are correct that it does happen to some extent, but I didn't ask whether it happens at all, but whether the current setup yields "regularly and effectively" as to support your preposition that the community does indeed "decide how many pearls a player should hold?" And at the very least from my own perspective, given that my voice, as part of the "community", has never, not once, been sought after nor listened to.. I cannot see how the *community* decides these things and not the inner circles of alliances.

KaloHG:
I support this change tbh. If an exiled player is willing enough to log in to keep their streak going for as long as is needed to actually be "forcibly" freed then they kinda deserve to play. It just doesn't make sense to me that players can be pearled over petty shit, keep playing for weeks or even months after, and hardly make a dent in built up nations essence costs. I don't believe this change will cause extremely important pearls to be freed, as most "important" pearls are pearls of people who are banned, inactive, or they don't login enough to be freed via streak.

Another consideration is the major concern from WP is "oh nooo they can all login and get freed oh nooo". Well the truth is, it doesn't really matter. These people have dug their heels into this map for over 3 years nearly, and at this point there is no reasonable or possible way for anyone to contest them without that being permitted by themselves. You can just easily repearl the people you pearled if they're that much of a bother to you. For example, PapaPound is someone acknowledged to be a POS individual in every single nation on the map. If they get freed due to pearl costs, it doesn't really matter because they'll be pearled **EXTREMELY** quickly.

Diet-Cola:
> If an exiled player is willing enough to log in to keep their streak going for as long as is needed to actually be "forcibly" freed then they kinda deserve to play.

This is what the current system enforces, spiralling costs for those who log in and maintain a streak and every 270 days they double up in cost

> It just doesn't make sense to me that players can be pearled over petty shit, keep playing for weeks or even months after, and hardly make a dent in built up nations essence costs.

This is a community issue and not a balance issue with pearl costs, if you believe someone is held unjustly its your job to kick up a fuss to get them freed. Not the admins decision to ramp up fuel costs so you can drain your captors essence bank which they have only acquired from logging in and being active which is the whole argument that "pearled players don't play cuz pearled"

> oh nooo they can all login and get freed oh nooo

Logging in and maintaining a streak is easier than ever on classics, 2.0 had prison pearl and 8 coal per day costs, never an issue there. Exile pearl has made being pearled more playable than ever, the people who get pearled and don't log in anymore and whine to increase pearl costs are usually the people who are unwilling to cooperate and hence where they've ended up.

> You can just easily repearl the people you pearled if they're that much of a bother to you.

1. This is extremely easier said than done
2. Welcome to repearl craft

Overall, past iterations of increasing pearl costs has only resulted in driving away the core people who are keeping the server active. If every pearl was freed out of the hjalt vault tomorrow, do you actually believe their would be a spike in activity? 

And as for permanent charge fee, as bad as it might sound that some people have been pearled for 3 years, those people have either never logged in since their pearling or never done anything to amend their pearling. So now we want to go out of our way to cater the server towards these people at the expense of the active player base who regularly login? This is completely backwards.

Being pearled sucks, it's supposed to be that way. We seriously need to deviate from having this debate every 3 months since it always always always ends up getting incrementally more expensive and that will push away people playing to fuel them.

KaloHG:
I honestly think pearl costs are perfectly fine as is. there's no reason to modify them. I mean an increase would be... interesting to say the least. But lowering them wouldn't really make sense, as currently clearly pearls that need to be held are able to be held just fine.

PR #200, opened 2021-06-02
Adding new 1.16 blocks + brewing stands to chunk limits.

Convoy20:
Adding new 1.16 blocks + brewing stands to chunk limits.
If you do not merge this I will make blast furnace land. This is not a threat, it is a promise.

### DIFF
--- a/templates/public/plugins/SimpleAdminHacks/config.yml.j2
+++ b/templates/public/plugins/SimpleAdminHacks/config.yml.j2
@@ -310,6 +310,11 @@ hacks:
         DROPPER: 16
         COMPARATOR: 12
         FURNACE: 40
+        BLAST_FURNACE: 40
+        BREWING_STAND: 40
+        SMOKER: 40
+        CAMPFIRE: 20
+        SOUL_CAMPFIRE: 20
       exempt:
       - Steve
       - Alex
--- a/templates/public/plugins/SimpleAdminHacks/config.yml.j2
+++ b/templates/public/plugins/SimpleAdminHacks/config.yml.j2
@@ -313,8 +313,9 @@ hacks:
         BLAST_FURNACE: 40
         BREWING_STAND: 40
         SMOKER: 40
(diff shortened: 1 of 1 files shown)


Ameliorate:
Campfires might be a slight issue. Extinguished campfires make quite the nice building block.

Convoy20:
> Campfires might be a slight issue. Extinguished campfires make quite the nice building block.

I'm dubious if this should be kept because of how campfire's models are multiblock I think they're way laggier than signs. Could bump the limit though.

Convoy20:
Maybe we should include armor stands into this too?

wingzero54:
> 
> 
> Maybe we should include armor stands into this too?

Would that conflict with the gui's which use invisible armor stands?

wingzero54:
Also campfires ticked 692,813 times but only used 0.04% of the server, so I think we can safely leave them at 40

Convoy20:
> > Maybe we should include armor stands into this too?
> 
> Would that conflict with the gui's which use invisible armor stands?

No, it should only be cancelling player interactions

Diet-Cola:
Blocking armour stands probably wont work since theyre fully an entity, not a TileEntity

Diet-Cola:
Plus it is way too easy to work around a chunk limit for armour stands

Convoy20:
> Blocking armour stands probably wont work since theyre fully an entity, not a TileEntity

Can you check, don't know what it does to stop them from being placed.

wingzero54:
Tested and confirmed the armor stand limit does nothing. This hack is a limit on block placement, but the armor stand becomes an entity when placed. I don't think the limit would ever come into play

Convoy20:
> Tested and confirmed the armor stand limit does nothing. This hack is a limit on block placement, but the armor stand becomes an entity when placed. I don't think the limit would ever come into play

Removed armor stands from the config - should be good to go now

PR #201, opened 2021-06-03
Disable phantoms.

Convoy20:
No one likes them. It hurts retention.

### DIFF
--- a/templates/public/plugins/SimpleAdminHacks/config.yml.j2
+++ b/templates/public/plugins/SimpleAdminHacks/config.yml.j2
@@ -259,7 +259,7 @@ hacks:
     villagerTrading: false
     witherSpawning: false
     patrolSpawning: false
-    phantomSpawning: true
+    phantomSpawning: false
     enderChestPlacement: true
     enderChestUse: false
     shulkerBoxUse: false
--- a/templates/public/plugins/SimpleAdminHacks/config.yml.j2
+++ b/templates/public/plugins/SimpleAdminHacks/config.yml.j2
@@ -167,7 +167,7 @@ hacks:
     # they will start spawning phantoms. Setting the cap to below 72k will effectively disable phantoms. Setting the
     # cap to -1 will disable the cap. Default: 80k (10% chance per two minutes during the night and thunder storms.)
     # Read more here: https://minecraft.gamepedia.com/Phantom#Java_Edition
-    timeSinceRestCap: 75000
+    timeSinceRestCap: 0
     # Determines whether phantoms are allowed to spawn during the night.
     nightSpawn: true
     # Determines whether phantoms are allowed to spawn during thunder storms.


Protonull:
Not true. Proof?

Convoy20:
No one who actually plays likes them*

Protonull:
Still not true.

Convoy20:
https://i.imgur.com/GJ2zHE4.png

Protonull:
seven people, much wow :)

xFier:
I only log in accidentally, among other things, I blame the phantoms

notdrop:
If you're so defensive of the fact that people like them, I highly recommend polling on discord to get a look into community opinion. This mechanic is pointless other than to antagonize people with.

I am almost willing to bet the entirety of my wealth in game that you'll get an overwhelming majority _against_ phantoms.

xFier:
> seven people, much wow :)

Find me seven people who do like them…

Protonull:
> > seven people, much wow :)
> 
> Find me seven people who do like them…

As mentioned in-game, few are genuinely "pro"-phantom, instead they're more anti-anti-phantom with the attitude of "what's the big deal" and "they have uses in factories and such" and "it's a vanilla feature." Just because you can find rabid support for a cause, doesn't mean the opposing side needs to be equally as rabid.

Convoy20:
I think the only productive thing from in-game chat argument about the issue, is that people find phantoms annoying and more importantly they don't drop anything actually interesting to keep their annoyingness. 

If this PR isn't accepted, I think a compromise would be to find either more drops for phantoms to give or something that makes it exciting if they spawn, rather than a hassle.

Convoy20:
Another compromise would be to spawn them but with no aggro: /summon minecraft:phantom ~ ~1 ~ {Attributes:[{Name:"generic.followRange",Base:0}]}

Protonull:
Making them a passive mob is a novel idea, though if this were to happen I'd prefer being them a neutral mob instead. (Passive being like pigs and chickens, and neutral being like bees and endermen)

Ameliorate:
Another option is to make them spawn only in hell biomes. This keeps them in the game while removing them as an annoyance feature.

Protonull:
> Another option is to make them spawn only in hell biomes. This keeps them in the game while removing them as an annoyance feature.

Hmmm, in that scenario it might be best to detach phantoms from insomnia entirely and have them spawn like regular mobs but just in hell. Thoughts?

Convoy20:
> > Another option is to make them spawn only in hell biomes. This keeps them in the game while removing them as an annoyance feature.
> 
> Hmmm, in that scenario it might be best to detach phantoms from insomnia entirely and have them spawn like regular mobs but just in hell. Thoughts?

Probably the easiest solution, don't know how easy it'd be to change their mob-ai faction

Protonull:
> > seven people, much wow :)
> 
> Find me seven people who do like them…

Can't name them, but this [Reddit Poll](https://www.reddit.com/r/civclassics/comments/nrlotd/do_you_like_phatoms/) indicate 26 persons, 1 including myself of course, that like them or at the very least don't see a reason to remove them. This is still a minority opinion (if the poll is taken at absolute fact and there's no foul play as indicated by anti-phantomers who've said they'd use reddit-alts to boost their numbers), but it still exposes as false this narrative that anti-phantom sentiments were nigh on unanimous and I was the only person who liked them.

caucow:
> I live phantomes
by someone with "monkeying around" as a title... so a civ player
> making reddit alts to say no
by gregy who's eternally an antagonist and shitposter and probably didn't bother with the effort
> why would anyone dislike phatoms
can't tell if sarcasm
> Love phantoms, they’re just the best
definitely sarcasm
>binary poll for non-binary question
your poll is bad and your polling platform is bad. you have no user verification, minimal user profiling and no nuance to the variants of "yes" or "don't care" such as the aforementioned anti-anti-phantomers or orange wizard's based af and relatable response
> i don't like phantoms but i like the pain and suffering they cause to other people

caucow:
as much as it is my general opinion that people are bad and should feel bad, that's not what I'm saying here  
if you have a poll that is going to have some administrative decision tied to it, attach names to the results and provide usable nuanced answers rather than just leaving it up to shitposters on the anonymous internet to pick a binary response  
*and* make it known that there is some actual administrative weight to it, you'll probably get more+slightly better results

wingzero54:
@Protonull  is converting Phantoms to passive ai something your mob ai hack is capable of doing? You could maybe just make them never aggro anything but still keep spawn conditions?

Protonull:
> @Protonull is converting Phantoms to passive ai something your mob ai hack is capable of doing? You could maybe just make them never aggro anything but still keep spawn conditions?

Unfortunately no, that hack is for outright disabling AI, so the phantoms would just remain stationary in the air without regard to anything. You wouldn't even be able to move them by running into them or hitting them.

wingzero54:
Sorry but not gonna happen

PR #203, opened 2021-06-08
Drastically increase voting rewards

Convoy20:
Daily login shouldn't be nearly 3x voting rewards at full streak.

5 per site gives a max daily essence reward of 23, compared to current 11 essence max.

But let's be real, few people will vote, so average max across players will be 15 but this gives people some actual incentive to vote.

### DIFF
--- a/templates/public/plugins/EssenceGlue/config.yml.j2
+++ b/templates/public/plugins/EssenceGlue/config.yml.j2
@@ -21,7 +21,7 @@ voting_reward:
   essence:
      material: ENDER_EYE
      name: Player Essence
-     amount: 1
+     amount: 5
      lore:
       - Activity reward used to fuel pearls
 


Protonull:
While I can understand wanting to increase the incentive to vote.. I don't think making voting twice as lucrative as a *maxed out* activity-streak is really going to help things. Voting should imo remain supplemental.

Convoy20:
We're at a little under 100 votes across 7 days in this month - 14 people voting on a 70 cap server is pretty bad. Equivalent vote count servers are like 2-10 players at night, compared to our 20-30. 

Incentives are not there. If not essence, what else would you give @Protonull

Protonull:
Coloured names, tab flairs, etc.. if we set aside some time to think about it, we could probably come up with quite a few things we can reward players with that do not wildly inflate the amount of essence in the world, effectively making pearl costs negligible.

Convoy20:
> Coloured names, tab flairs, etc.. if we set aside some time to think about it, we could probably come up with quite a few things we can reward players with that do not wildly inflate the amount of essence in the world, effectively making pearl costs negligible.

I mean, part of this increase would also include me advocating for either making essence more useful for other things (thereby making pearl costs more expensive *gasp*)  or worse option making pearl costs marginally more expensive to compensate. 

I asked in ! chat what voting rewards were and literally laughed at it. So something quickly needs to change because it's entirely off base right now.

Protonull:
Having someone vote and receive twice more essence than if they spent five days in a row being active, or two days at a maxed out streak, is a disruption to the essence economy so large that it cannot just be fixed by adding pearl costs to this or that.. like, there are already discussions about adding pearl costs to things within the *current* essence economy. People are already building massive stocks of essence because unless you're a pearl hoarder, that essence has nowhere to go.. and all this change will do is triple the size of those stocks.

And while I agree that changes need to be made to voting rewards, we should try and add cosmetics into the equation rather than just headfirst into hyperinflation.

Convoy20:
> Having someone vote and receive the same amount of essence as spending five days in a row being active

People who vote are worth waaaay more than people who login (tbh usually through a bot)

wingzero54:
I won't accept this change, I don't think voting is worth that much and inflating essence doesn't match with long-term plans

PR #204, opened 2021-06-08
Reduce Cobble->Gravel and Gravel->Sand outputs

Protonull:
Given the vast impact this is having on sand and gravel markets, the factory exchange rate between cobble, gravel, and sand is too even.. and charcoal costs are irrelevant given how incredibly cheap and bottable wood is.


wingzero54:
Closed for #210

PR #206, opened 2021-06-08
Nerf to Grinding mill

Lodish:
A different nerf to the grinding mill, adding flint to the mix

### DIFF
--- a/templates/public/plugins/FactoryMod/config.yml.j2
+++ b/templates/public/plugins/FactoryMod/config.yml.j2
@@ -5725,34 +5725,43 @@ recipes:
       stone:
         material: STONE
         amount: 32
+      flint:
+        material: FLINT
+        amount: 1
     output:
       cobblestone:
         material: COBBLESTONE
         amount: 32
   crush_cobble:
-    production_time: 4s
+    production_time: 16s
     name: Crush Cobblestone
     type: PRODUCTION
     input:
       cobble:
         material: COBBLESTONE
-        amount: 32
+        amount: 64
+      flint:
+        material: FLINT
+        amount: 4
     output:
       gravel:
         material: GRAVEL
-        amount: 32
+        amount: 48
   crush_gravel:
-    production_time: 4s
+    production_time: 16s
     name: Crush Gravel
     type: PRODUCTION
     input:
       gravel:
         material: GRAVEL
-        amount: 32
+        amount: 64
+      flint:
+        material: FLINT
+        amount: 4
     output:
       sand:
         material: SAND
-        amount: 32
+        amount: 48
   grind_sandstone:
     production_time: 4s
     name: Grind Sandstone
@@ -5761,6 +5770,9 @@ recipes:
       sandstone:
         material: SANDSTONE
         amount: 64
+      flint:
+        material: FLINT
+        amount: 1
     output:
       sand:
         material: SAND
@@ -5773,6 +5785,9 @@ recipes:
       red sandstone:
         material: RED_SANDSTONE
         amount: 64
+      flint:
+        material: FLINT
+        amount: 1
     output:
       red sand:
         material: SAND
@@ -5787,6 +5802,9 @@ recipes:
         amount: 64
       leather:
         material: LEATHER
+        amount: 4
+      flint:
+        material: FLINT
         amount: 4
     health_gained: 10000      
 


Ameliorate:
This is very similar to #205. It might be worth explaining how this PR is different from the other Grinding Mill nerfs.

Protonull:
The flint rates are different, it also doesn't include the recipe to turn coarse dirt into dirt, but includes a vast increase to the production time.

Lodish:
#205  increases a flat rate of 8 flint to every recipe, this one adds 4 flint to the stone>cobble>gravel>sand ones and 1 flint to some of the others

it also changes the amounts so the conversion isn't 1:1 but 4:3 for each step, and the charcoal cost (aka production time) is 4x higher

#205 also adds flint to the factory creation and repair recipes which I should have done

Quinndolin:
Both of these nerfs are severe enough that they pretty much make the recipes useless. Why have them at all at this point? Allowing people to use a reasonable amount of iron or diamond as a 'grinding ball' would be a better choice than this amount of flint. Enables people to sell hand collected sand to xp producers at a rate lower than the fixed grinding mill one.

Protonull:
> Both of these nerfs are severe enough that they pretty much make the recipes useless.

Eh, not really. Both cobblestone and charcoal are renewable resources. You'll just need to use more of them to create vast stockpiles now.

Quinndolin:
> Eh, not really. Both cobblestone and charcoal are renewable resources. You'll just need to use more of them to create vast stockpiles now.

I'm talking about the flint costs, cobblestone and charcoal are completely irrelevent and are basically infinite.

Protonull:
But why does the inclusion of a non-renewable resource\* mean that the recipes are useless? That kind of implicitly suggests that the recipes are only there *in order to* produce infinite amounts of gravel and sand on the cheap.

[\*] Flint will technically be renewable given that gravel is renewable, and you primarily get flint from gravel, but the drop rates and the reduced production means you can't create infinite sand from infinite gravel.

Quinndolin:
> 
> 
> But why does the inclusion of a non-renewable resource* mean that the recipes are useless? That kind of implicitly suggests that the recipes are only there _in order to_ produce infinite amounts of gravel and sand on the cheap.
> 
> [*] Flint will technically be renewable given that gravel is renewable, and you primarily get flint from gravel, but the drop rates and the reduced production means you can't create infinite sand from infinite gravel.

You're putting words in my mouth. The inclusion of flint at this amount makes the recipe useless for anything other than producing very small amounts of sand. At that point why not just remove them? The only use case for such a small amount of sand is modest builds, which people already don't use these recipes for anyway. If any of these prs goes through these recipes will literally just not get ran.

Protonull:
I mean, I did say "kind of implicitly suggests" :P

Quinndolin:
> 
> 
> I mean, I did say "kind of implicitly suggests" :P

>than this amount of flint

I literally cite the amount as being the problem and you respond with
>But why does the inclusion of a non-renewable resource* mean that the recipes are useless? 

"The inclusion of a non-renewable resource" isn't implicitly suggesting **anything** because I didn't say that

Protonull:
Since cobble and charcoal are a renewable resource and thus, as you say, are completely irrelevant to the equation, the addition of flint as a scarce resource (really anything can be used, but it'd be nice to give flint a use) is the only way to really balance out the Grinding Mill short of just removing it. I don't think requiring 8 flint is too bad for a 4:1 ration recipe, though perhaps the input:output can be changed from 32:8 to 64:16, or 128:32, etc.. either way, the addition of the flint is *supposed* to mean that you actually need to collect the scarce resource to run the recipe instead of just totally relying on the factory.

Quinndolin:
If you want to prevent a glass printer the amount of gravel produced per run needs to be less than the flint cost. F3 Shovel will always drop a flint and can be offhanded to reduce repair costs.

>I don't think requiring 8 flint is too bad for a 4:1 ratio recipe, though perhaps the input:output can be changed from 32:8 to 64:16, or 128:32

The inconveniance of having to collect flint will mean this recipe does not get ran for small amounts of sand. The huge amount of flint needed for 128 compacted glass bottles will mean this recipe does not get ran for xp (unless the above condition isn't met). With the recipe fitting neither one of these use cases why not just remove it? The charcoal cost is also problematic for most people in use case number 1.

A modest iron cost makes use case number 1 more accessible as people will have iron lying around. It preserves the role grinding mills can play in xp production if people are willing to spend iron instead of time spent mining sand, as I said it also presents the opportunity for people to sell sand to xp producers.

Protonull:
> The huge amount of flint needed for 128 compacted glass bottles will mean this recipe does not get ran for xp

Good tbh

wingzero54:
Closed for #210

PR #207, opened 2021-06-08
black dye removal

Lodish:
Remove black dye recipe, it was added because squids were not spawning, and the recipes addition nerfed a niche market. squids are now plentiful.

### DIFF
--- a/templates/public/plugins/FactoryMod/config.yml.j2
+++ b/templates/public/plugins/FactoryMod/config.yml.j2
@@ -53,7 +53,6 @@ factories:
      - charcoal_from_acacia_log
      - charcoal_from_dark_oak_log
      - charcoal_from_coal
-     - lamp_black
      - repair_charcoal_factory
   aesthetics:
     type: FCC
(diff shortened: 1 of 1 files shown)


Ameliorate:
While I stated the recipe's reason for being added in #feedback, I don't give support for removing the recipe.

I support keeping the recipe in the game, because black dye is some of the hardest dye to get in the game. Cheap dyes mean that map art is easier to make.

In addition, this recipe gives black dye. However, some recipes require specifically squid ink, which you still have to farm by killing squids.

Lodish:
> In addition, this recipe gives black dye. However, some recipes require specifically squid ink, which you still have to farm by killing squids.

the only recipe is making book and quills, removing the supply of ink (because less people will bother to farm and sell them) just makes literature more expensive

Ameliorate:
Copying books using a printing plate also uses squid ink.

AllenY:
strongly support; black dye is among the hardest dye in the game to get, but it is still very easy to get; it is possible to afk farm it, but even if you morally object to afk farming, you can get like a stack in 15 minutes manually, and we of course already have massive amplification of the power of dye in factories. the game having no economic scarcity is a bad thing, especially for things such as art and aesthetic builds which are luxuries, and which will only stimulate the economy if they are not trivially easy to get resources for. as an aside I think the recipe is a bad idea because mining sucks

Protonull:
I also support removing this recipe.. as someone who's only just recently done a few runs of the Printing Press and had to go Ink Sac farming, it was literally like.. 2-3 minutes? Just went to the nearest river biome and wait for about half a minute for a squad to spawn, takes about another half minute to kill them all and by that time another squad has spawned. *(Side note, swimming to kill squids did make me realise what the clickers meant by swim combat being aids)*

wingzero54:
The original reason for adding this is because there was a couple month period where mobs were turned off entirely, so ink sacs were impossible to farm. But that is no longer the case so I agree with removing this recipe

PR #209, opened 2021-06-08
[1.17] Disable "bugged" obsidian generation

Protonull:
So the whole string to obsidian generator thing is, iirc, a bug that became a feature to help with bulk obsidian generation. I don't know the actual history so that's my abridged version. *Anyway,* in 1.17 you can use dripstones below a lava source to drip lava into a cauldron to produce what I assume to be infinite lava, thus bulk obsidian generation can return to a more vanilla, minecrafty way that also happens to be more accessible to newfriends.

### DIFF
--- a/templates/public/plugins/SimpleAdminHacks/config.yml.j2
+++ b/templates/public/plugins/SimpleAdminHacks/config.yml.j2
@@ -272,7 +272,7 @@ hacks:
       playerKill: 5
     blockWaterInHell: true
     minecartTeleport: true
-    obsidianGenerators: true
+    obsidianGenerators: false
     personalDeathMessages: false
     disableNetheriteCrafting: false
     goldBlockTeleport: false


Protonull:
[See here](https://youtu.be/4yieNs8mwIs?t=947) for a demonstration

Ameliorate:
https://www.youtube.com/watch?v=AdoX3m2I4tw is a design for a farm based on the vanilla idea.

Protonull:
> https://www.youtube.com/watch?v=AdoX3m2I4tw is a design for a farm based on the vanilla idea.

Looks good :)

EnforcerFifteen:
Why not leave both functions as generation methods? Unless this causes some terrible bug on update I think multiple ways to do the same process breeds different designs instead of just pointing players in one min-max direction.

wingzero54:
I agree with Enforcer, I'd be interested in seeing the new method brought in and used (eventually) before we look at disabling the old method

MrJeremyFisher:
I think currently, generating obsidian through dripstone is impossible. If I'm not wrong, the server would either need a new world with 1.17 generation or factories would need to be edited to facilitate the creation of dripstone.

Protonull:
> I think currently, generating obsidian through dripstone is impossible. If I'm not wrong, the server would either need a new world with 1.17 generation or factories would need to be edited to facilitate the creation of dripstone.

Hence the `[1.17]` tag :P

smaldragon:
> I think currently, generating obsidian through dripstone is impossible. If I'm not wrong, the server would either need a new world with 1.17 generation or factories would need to be edited to facilitate the creation of dripstone.

Pointed dripstone grows as a crop and allows you to craft further dripstone blocks, as long as there is a bio-lab like recipe for obtaining the first few needed to kickstart a farm that shouldn't be an issue

KaloHG:
ok lets continue to ignore all of the valid arguments that were made by numerous individuals in #feedback as well as in just general vc's and attempt to force an obscure vanilla method that makes creating obsidian far more annoying. 

thank you orinnari, my day has officially been thoroughly ruined.

Protonull:
Your day has been ruined because I restored an accidentally deleted branch? Yeesh.

EvilAsuramaru:
Why does this exist in the first place?

Protonull:
Read the first comment?

EvilAsuramaru:
I did? It didn't help me figure out why dripstone/lava is better than string or why this would be more available to newfriends.

Protonull:
Because I think there's value in vanilla mechanics. I'm **NOT** saying that custom mechanics are bad, but that, *generally speaking*, custom mechanics should extend vanilla mechanics, not replace them. Take CivEx for example: I've been unreservedly critical of its item-repair and potion mechanics mostly because they were entirely custom (as in that they also removed their vanilla counterparts) and thus were confusion and off-putting. The admins considered this a mere documentation problem, that they just need to put enough explanatory text as lore and everything'll be fine, and while I think has *some* merit to it, I think those mechanics were predisposed to failure given their divergence from and replacement of vanilla mechanics. But I digress.

And so, after the 1.17 update we'll have a situation of having two means of producing infinite obsidian. I personally think the vanilla mechanic should prevail, as it were, but I *am* content with Enforcer's idea of just leaving them both. I'm not fervently pursing this.. it's been over 3 months since the last message and the only reason we're talking now is that I accidentally pruned the "patch-3" branch and restored it. If Wing decided to close this PR, I'll be content with that, but I conscientiously object to this notion around that I should forever close this PR in shame because it's controversial.

So, to answer your questions directly and explicitly: it's more ~~available~~ accessible to newfriends by virtue of it probably being a mechanic they're already familiar with. And I never said that string or drip generation is better or worse than the other, I said one is vanilla and one is a custom bug-turned-feature.

Doommad:
How many people that are new friends even play 1.17. Like this whole discussion of it being more accessible to new friends is based on them already being familiar with the mechanic. I know many people that prefer modded servers which mechanic are built off of 1.16.4 and earlier that also include a fair amount of custom plugin. At what point does the stuff that is modified or custom get so diluted down that people choose a vanilla server over Civclassic.

Protonull:
This whole discussion of it being more accessible to newfriends is secondary.. it was a "oh and also" at the end of the first comment, it's not actually the reason for the change. I don't really understand the argument though.. are you suggesting that players come to Civclassics because string->obsidian is a custom mechanic and that's neat? And so removing it is going to cause people to go flooding towards vanilla servers? Like, what?

PR #210, opened 2021-06-08
Updated Grinding Mill Nerf (I understand github a bit better)

Lodish:
added flint to repair and factory creation

### DIFF
--- a/templates/public/plugins/FactoryMod/config.yml.j2
+++ b/templates/public/plugins/FactoryMod/config.yml.j2
@@ -5725,34 +5725,43 @@ recipes:
       stone:
         material: STONE
         amount: 32
+      flint:
+        material: FLINT
+        amount: 1
     output:
       cobblestone:
         material: COBBLESTONE
         amount: 32
   crush_cobble:
-    production_time: 4s
+    production_time: 16s
     name: Crush Cobblestone
     type: PRODUCTION
     input:
       cobble:
         material: COBBLESTONE
-        amount: 32
+        amount: 64
+      flint:
+        material: FLINT
+        amount: 4
     output:
       gravel:
         material: GRAVEL
-        amount: 32
+        amount: 48
   crush_gravel:
-    production_time: 4s
+    production_time: 16s
     name: Crush Gravel
     type: PRODUCTION
     input:
       gravel:
         material: GRAVEL
-        amount: 32
+        amount: 64
+      flint:
+        material: FLINT
+        amount: 4
     output:
       sand:
         material: SAND
-        amount: 32
+        amount: 48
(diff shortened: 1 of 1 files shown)


Lodish:
this is #206 with added factory creation costs from one of the other suggested nerfs

wingzero54:
I like the flint costs of this PR, but I think a better option of sand cost would be a variation of #204 . I like this general idea:
64 Stone -> 48 Cobble
64 Cobble -> 32 Gravel
64 Gravel -> 16 Sand

Where you step down in efficiency as you go, instead of simply raising charcoal cost. Thoughts?

Lodish:
ok I updated it:
removed the charcoal increases
changed the input to output ratios
64 stone +4 flint -> 48 cobble
64 cobble +4 flint -> 32 gravel
64 gravel +4 flint -> 16 sand

Lodish:
halved the amounts but maintained the ratio
32 stone +2 flint -> 24 cobble
32 cobble +2 flint -> 16 gravel
32 gravel +2 flint -> 8 sand

civrealmsmap:
I know this has been said at the previous PR already so I might miss some explanations or discussion, but what exactly is the point of keeping a rate of 64 cobble+6 flint to 8 sand, who is expected to still use it.
The smaller XP actors and builders will find this too much of a hassle and cost so traditional sand mining will be needed, and the bigger groups will, in anticipation of this, simply stockpile immense amounts of sand already and can treat these resources closer to infinite as well (which isn't a problem as that is part of the point of being a bigger/more effective group).
So what exactly is being achieved here, what problem is being solved?

Protonull:
Stockpiles are, at least in my opinion, the least appropriate argument against balance changes like this. It's true that people will have sand stockpiles and are perhaps making them right now due to the nature of open source and open feedback, but using this as a reason to *not* balance things is only going to increase the number and size of those stockpiles. Once this change is accepted, those stockpiles start being tapped into and will eventually run dry. It might take a while but the result is inevitable. If we refrained from balancing something because the effect wont be felt immediately by everyone in the same way, we'll never be able to balance anything.

civrealmsmap:
The stockpile part is rather minor though, it should be obvious to see that the one's profitting of the change, simply by facing the least amount of disturbance through it, are the big XP groups, and the ones facing most suffering are primarily smaller to mid building groups, or if you will the small XP groups. The question is, does this align with whatever problem is to be addressed here, is the goal to buff big nations and to nerf small ones and builders? I think the sheer utility of the base recipe for non-XP-activity has to be seriously considered. Further, changing this recipe in this way feels slightly disingenous, as effectively gutting it would leave calling for direct removal as the more honest approach

PR #214, opened 2021-06-17
Grinding Mill Buff

yodabird19:
The Grinding Mill is a blessing for the oldfriend and the newfriend alike. All this talk of nerfing the Mill is nonsense, how about we buff it instead?
- Added Crush Compacted Stone
- Added Crush Compacted Cobble
- Added Crush Compacted Gravel
- Added Grind Compacted Sandstone
- Added Grind Compacted Red Sandstone
- Added Crush Compacted Coarse Dirt
- Added Shave Compacted Snow From Compacted Ice

### DIFF
--- a/templates/public/plugins/FactoryMod/config.yml.j2
+++ b/templates/public/plugins/FactoryMod/config.yml.j2
@@ -838,6 +838,13 @@ factories:
      - grind_red_sandstone
      - crush_coarse
      - shave_snow
+     - crush_compacted_stone
+     - crush_compacted_cobble
+     - crush_compacted_gravel
+     - grind_compacted_sandstone
+     - grind_compacted_red_sandstone
+     - crush_compacted_coarse
+     - shave_compacted_snow
      - repair_grinding_mill
 #RECIPES
 recipes:
@@ -5897,6 +5904,22 @@ recipes:
       dirt:
         material: DIRT
         amount: 64
+  crush_compacted_coarse:
+    production_time: 4s
+    name: Crush Compacted Coarse Dirt
+    type: PRODUCTION
+    input:
+      coarse_dirt:
+        material: COARSE_DIRT
+        amount: 64
+        lore:
+          - Compacted Item
+    output:
+      dirt:
+        material: DIRT
+        amount: 64
+        lore:
+          - Compacted Item
   shave_snow:
     production_time: 4s
     name: Shave Snow from Ice
@@ -5909,6 +5932,22 @@ recipes:
       snow:
         material: SNOW_BLOCK
         amount: 64
+  shave_compacted_snow:
+    production_time: 4s
+    name: Shave Compacted Snow from Compacted Ice
+    type: PRODUCTION
+    input:
+      ice:
+        material: ICE
+        amount: 64
+        lore:
+          - Compacted Item
+    output:
+      snow:
+        material: SNOW_BLOCK
+        amount: 64
+        lore:
+          - Compacted Item
   crush_stone:
     production_time: 16s
     name: Crush Stone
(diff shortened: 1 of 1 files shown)


SpacemanSpleef:
Too underpowered

Metriximor:
This is sorely needed

Building with sandstone is a nightmare cause it's so expensive

Quinndolin:
Perhaps a stone -> sandstone recipe path that's much more generous but doesn't let you get sand for glass?

Protonull:
> Building with sandstone is a nightmare cause it's so expensive

image

yodabird19:
A stone->sandstone recipe would certainly be efficient, but let's not discount the other three uses of sand?
- concrete
- glass
- itself

Protonull:
Concrete would grind back into concrete powder, not sand.

yodabird19:
What? No, I'm listing _uses_, not sources; my point is that sand being accessible isn't a bad thing.

InvertedMind0:
based

Metriximor:
> Perhaps a stone -> sandstone recipe path that's much more generous but doesn't let you get sand for glass?

That would be good yes, maybe in the aesthetic factory

Quinndolin:
> 
> 
> > Building with sandstone is a nightmare cause it's so expensive
> 
> image

Crafting sandstone is probably the most expensive conventional building material. Not sure about how much actually generates or how feasible it is to collect.

Protonull:
Not everything should be available to everyone anywhere cheaply and easily.. and even then, deserts aren't rare, neither are beaches, rivers, or oceans. Sandstone generates naturally with like 2-5-ish layers below the sand for the entire biome. You can also craft sandstone from sand, which is infinitely generate-able from the Grinding Mill.

yodabird19:
Certainly not everything, but what good comes from an _exclusively_ building material being such a pain to get?
I suppose this PR would go a good way to fixing that, seeing as it makes generating that crafting sand a whole lot quicker.

Protonull:
> Certainly not everything, but what good comes from an _exclusively_ building material being such a pain to get?

Because it wouldn't be "*exclusively*" a building material.. it would also be a means to bypass the suggested nerfs to `Stone->Gravel->Sand` rates.

yodabird19:
Based then

Metriximor:
> > Certainly not everything, but what good comes from an _exclusively_ building material being such a pain to get?
> 
> Because it wouldn't be "*exclusively*" a building material.. it would also be a means to bypass the suggested nerfs to `Stone->Gravel->Sand` rates.

But I'm not trying to generate sand, I'm trying to make sandstone which afaik is an end product. The stone gravel sand route is extremely expensive(for good reason since glass is used in exp production)

As for the exclusivity of the material, I disagree, the existence of the aesthetic factory shows that material exclusivity is not the intent of civ, and also deserts are not nearly common enough.

yodabird19:
The argument that sandstone access is bad because it means greater access to sand requires that everyone agree that greater access to sand is a bad thing; I think you'll find they don't.

templates/public/plugins/FactoryMod/config.yml.j2
```diff
@@ -5897,6 +5904,22 @@ recipes:
       dirt:
         material: DIRT
         amount: 64
+  crush_compacted_coarse:
+    production_time: 4s
```
Convoy20, line 5908: What doesn't make sense in this request is that the production time is the same as normal production. Same charcoal cost for major increase in output.

I'd change this to be 256s for a full stack and 128s for a half stack to make charcoal costs the same.

wingzero54:
No

PR #215, opened 2021-06-17
Remove Uncarve [Compacted] Pumpkins

Protonull:
Since Wing's [commit](https://github.com/CivClassic/AnsibleSetup/commit/bd24ac3dd988dc07db1dcb5778dd712496d2c64c) that makes uncarving pumpkins 64x cheaper, this compacted variant should be removed. This recipe is singularly irregular, it's the only non-XP related recipe that processes compacted items, which fundamentally undermines the balancing of compaction; that XP recipes require compacted items because you simply cannot store that many items in a double chest, it's a necessity.. but uncarving pumpkins?

### DIFF
--- a/templates/public/plugins/FactoryMod/config.yml.j2
+++ b/templates/public/plugins/FactoryMod/config.yml.j2
@@ -716,7 +716,6 @@ factories:
      - adv_xp_5
      - adv_xp_6
      - uncarve_pumpkins
-     - uncarve_compacted_pumpkins
      - wither_skull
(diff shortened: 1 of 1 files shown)


wingzero54:
The purpose of the recipes were to convert pre-1.13 pumpkins to post-1.13 pumpkins, cost wasn't really a factor, it was just about people being able to convert existing stockpiles. I think at this point we can probably issue a notice and remove both recipes, they were really only meant to be temporary.

PR #219, opened 2021-06-26
Get logging for player spawn locations

Diet-Cola:


wingzero54:
Looks like debug just logs a player's spawn type, I don't think it logs the location of the spawn which is what I was looking for

PR #220, opened 2021-06-27
Add high piston limit to allow flying machines but also discourage ri…

Convoy20:
…diculous piston usage

Add high piston limit to allow flying machines but also discourage ridiculous piston usage

### DIFF
--- a/templates/public/plugins/SimpleAdminHacks/config.yml.j2
+++ b/templates/public/plugins/SimpleAdminHacks/config.yml.j2
@@ -325,10 +325,11 @@ hacks:
       exceededMessage: '&9Limit (%Limit%) reached for this chunk, you cannot place more of this type of block (%Material%).'
       tileEntities:
         HOPPER: 16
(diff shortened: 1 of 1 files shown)


wingzero54:
The flying machines I'm familiar with are 15 blocks in total (including pistons and every block touching/being moved by the pistons). Can you make any larger than that?

caucow:
> The flying machines I'm familiar with are 15 blocks in total (including pistons and every block touching/being moved by the pistons). Can you make any larger than that?

push limit is 12 blocks, so you can make a 24 block total machine. depending on design that can get to iirc 18 or 20 blocks long depending on how stable you want it to be during lag

Convoy20:
> The flying machines I'm familiar with are 15 blocks in total (including pistons and every block touching/being moved by the pistons). Can you make any larger than that?

I put it higher incase someone had multiple of those 15 block flying machines going at the same time. At ~30 if two of those machines entered the same chunk it could cause problems. 50 seemed high enough to avoid any issues which discouraging piston farms.

wingzero54:
Sure may as well I guess, as a failsafe

PR #223, opened 2021-07-02
Change how xp is made

Lodish:
An alternative to nerfing the grinding mill, xp requires a lored bottle, that itself needs essence and clay in it's creation

Input 192 glass, 16 clay balls, 1 essence
output 192 lored bottles required for xp recipes

### DIFF
--- a/templates/public/plugins/FactoryMod/config.yml.j2
+++ b/templates/public/plugins/FactoryMod/config.yml.j2
@@ -28,6 +28,7 @@ factories:
      - smelt_cracked_stone_bricks
      - smelt_glass
      - smelt_glass_red
+     - smelt_xp_bottle
      - smelt_netherrack
      - smelt_red_netherbrick
      - smelt_bricks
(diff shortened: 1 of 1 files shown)


Convoy20:
While I love wacky changes

doing this after producers have literally spent 3 years making XP and have pushed the price down to 1d per eblocks makes me think it's a little too late to change things

Lodish:
> While I love wacky changes
> 
> doing this after producers have literally spent 3 years making XP and have pushed the price down to 1d per eblocks makes me think it's a little too late to change things

Its more about fixing the grinding mill mistake, not really to increase xp cost

xFier:
This is a thinly-veiled attempt at pushing the same agenda as your grinder-nerf.

1.  You're attacking people's sand stockpiles in a manner that seems like you're trying to get back at them for stopping your previous attempt at this.

2. You could've added the clay and essence requirements directly to the XP recipes, but no, you want to make all of the glass bottles people have already made useless.

3. The essence requirement hurts everyone. An essence requirement only serves to put a cap on the amount of XP that can be produced in a single day, while hurting smaller groups who can now only produce a fraction of what bigger nations can.

EDIT: (after your admission that its more about fixing the grinding mill mistake) I think we're well past that, the "mistake" has happened and you only serve to make the game less enjoyable for the players.

Lodish:
> You're attacking people's sand stockpiles in a manner that seems like you're trying to get back at them for stopping your previous attempt at this.
 
This still requires sand so I'm not sure what you mean

> You could've added the clay and essence requirements directly to the XP recipes, but no, you want to make all of the glass bottles people have already made useless.

people can use glass bottles for other things, or just use them up before a change. 

> The essence requirement hurts everyone. An essence requirement only serves to put a cap on the amount of XP that can be produced in a single day, while hurting smaller groups who can now only produce a fraction of what bigger nations can.

It's one of the few unbottable things, what would you suggest instead?

xFier:
> It's one of the few unbottable things, what would you suggest instead?

The genie is out of the bottle at this point, you can't really replace something that used to be un-bottable (and scaled with how much effort someone was willing to put in) with something that everyone gets the same amount of regardless of effort.

yodabird19:
radical proposal: do nothing
every suggestion is getting more hate than not

Virizion:
> radical proposal: do nothing
> every suggestion is getting more hate than not

unironically this

dustcloud777:
> radical proposal: do nothing
> every suggestion is getting more hate than not

nah, keep the clay requirement but add it on top of everything else instead of changing bottles

PR #225, opened 2021-07-07
Fix grinding mill nerf cancellation

yodabird19:
I'm not quite sure what the problem was, but people have been reporting that the grinding mill nerf is still in effect despite its supposed cancellation 5 days ago; I've copied the exact configs for the seven grinding mill recipes from the last pre-nerf factorymod commit, so whatever was wrong, this should fix it.


wingzero54:
Issue was already resolved

PR #226, opened 2021-07-10
Make netherite pickaxes cost three times fewer netherite ingots

okx-code:
Currently you make 29% fewer net diamonds per time mining ancient debris in the hell biome with the netherite pickaxe, compared to the diamond pickaxe. This would make netherite pickaxes 1.4% better than diamond pickaxes.

Alternatively changing the ore bonus for netherite pickaxes to 45% would make them 2.2% better (50% is 7.3% better)


okx-code:
Netherite pickaxes are already better in the non-nether

PR #227, opened 2021-07-10
Increase rates of ancient debris in the nether biome

okx-code:
This would put the amount of money you make in a nether biome on par with what you make mining stone, and make netherite pickaxes 2.1% better to use, rather than 29% worse.

I would also slightly decrease the cost of repair (maybe the recipe for making netherite pickaxes makes 9 pickaxes instead of 8)

This would also change the nether biome from being five times worse to mine in to two times worse to mine in (when only counting diamond and netherite) per block break

### DIFF
--- a/templates/public/plugins/HiddenOre/config.yml.j2
+++ b/templates/public/plugins/HiddenOre/config.yml.j2
@@ -263,7 +263,7 @@ drops:
       transformIfAble: true
       biomes:
          NETHER_WASTES:
-            chance: 0.0018
+            chance: 0.006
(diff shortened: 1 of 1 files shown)


wingzero54:
This would make netherite 6x more common in nether biomes than regular overworld, I don't understand where your numbers are coming from

okx-code:
Yes, currently the amount of diamonds you get in the overworld still makes it about twice as worth it to mine there (assuming 3.5d per debris, but the maths works out up to 10d per debris). Therefore, I think it is still balanced

PR #229, opened 2021-07-18
Add in multiple layers of enderchests relay for shops

Convoy20:
I don't get why this is a setting, already coded, and not used?

It allows for actual wealth investment in shops, even if you relay a shop chest 3times, you still need to invest 12d in the glass, 3d in the enderchests, and then surround that relay with even more layers of DRO to make it worthwhile. (60d in obby on first layer, and you'd want each successive relay to have another additional layer of dro on it) A single secure relay system could *easily* cost 300d.

### DIFF
--- a/templates/public/plugins/ItemExchange/config.yml.j2
+++ b/templates/public/plugins/ItemExchange/config.yml.j2
@@ -83,7 +83,7 @@ shopRelay:
   relayBlocks:
     - "ENDER_CHEST"
   # Determines how many times a relay can be relayed
-  recursionLimit: 0
+  recursionLimit: 3
   # Determines how far (in blocks) a relay can search
   reachDistance: 4
   # Determines which blocks will be ignored by relays (wont end the search for shops)


Protonull:
Ganna relay concerns from the Discord here for posterity:

1. Doing this will break a lot of people's shops, like this:
image since all the relays will relay with each other. It's not an absolute deal-breaker, of course, but just be aware that any such change will be an immediate and genuine nuisance for quite a few people.

2. You are already able to drastically increase the protection of your shop chests via wealth sink, thus enabling relay recursion is simply unnecessary. We shouldn't just enable over-powered features because you want extraordinary protection while retaining a pretty facade. And even then, you can decorate and sign your shop facade just fine.

Convoy20:
> Ganna relay concerns from the Discord here for posterity:
> 
> 1. Doing this will break a lot of people's shops, like this:
>    image since all the relays will relay with each other. It's not an absolute deal-breaker, of course, but just be aware that any such change will be an immediate and genuine nuisance for quite a few people.
> 2. You are already able to drastically increase the protection of your shop chests via wealth sink, thus enabling relay recursion is simply unnecessary. We shouldn't just enable over-powered features because you want extraordinary protection while retaining a pretty facade. And even then, you can decorate and sign your shop facade just fine.

Pretty sure the relatively small amount of enderchest relay users (I'm thinking primarily ADH and Amel) would appreciate the small hassle of changing their shops for the added flexibility and protection they can then invest in. 

2: You really can't. It's exponentially bigger how much you can sink into a shop with even 1 or 2 more layers of recursion. 3 is what I picked in between the earlier discussions, but in the current config it'd be hard to spend more than 50 or so d per shop chest to defend.

Quinndolin:
Petty raiding is already too difficult for the high risk low reward activity it is.

Convoy20:
> Petty raiding is already too difficult for the high risk low reward activity it is.

Petty raiding isn't hitting shop chests, it's hitting new friends who put their diamonds in a stone reinforced chest.

Protonull:
YooooIheardYouLikedRelays

Convoy20:
Actually your 'basic relay design' is the most secure you can really efficiently make it. 

Break times: 
glass: 23.3 minutes
echest: 41.7 minutes
obby: 83.3 minutes

3.5 layers of glass per obby layer

If you do more than 1 layer of obby it's just easier to break the glass. Even obscuring the chest from view (which depending on how you do it can still be abused) At most you need to break two more slabs to get in which buys you 23 more minutes of break time, which is not worth another obby layer.

Protonull:
Iron bars: 40 minutes (though I acknowledge the risk of using bars, similar to the risk of using panes)

It's true that an abundance of obby will lead people towards weaker links, like the glass paths (for want of a better term), but I still do not understand why "not wanting obby boxes" or "they'll just break the glass" is reason for game-mechanical intervention.. is the goal really to make shop raiding impossible?

Indeed, this feels more like a criticism against Citadel as a whole. You want relays because the protection granted by bunkerifying (for want of a better term) the shop chest is not enough. So if Citadel is not good enough for shops, why is it good enough for homes and other storage?

okx-code:
Ori you're comparing breaking with no efficiency but convoy is comparing with eff 5

Convoy20:
I think iron bars would only work if you went straight down, which is a little weird but could work. And iron bars are 16.7 minutes, and since we're going straight down, now you only need to break one iron bar, so that's not more secure, at all.

>they'll just break the glass

I'm talking about efficiency. Not people blowing thousands of diamonds to flex about their obby abundance. Right now the total cost for a MAXIMALLY secure shop chest is 27 diamonds. For an additional like 5% protection buff we could up that cost to 
51 diamonds if you add the slab protections. 

No matter now if you spend more diamonds than on the 'basic relay shop' it's always faster as a raider to break the glass instead. Each additional layer of obsidian you put on it is useless. 

With additional relays it actually makes sense to do 2 or 3 layers of obsidian on your shop chest. 

>You want relays because the protection granted by bunkerifying (for want of a better term) the shop chest is not enough.

The issue with shop chests is two fold: 

1: The contents of the shopchest are broadcasted. If you had linked shopchests and you couldn't see where the item *actually* is you could do a chest vault to protect the items. 

2: The actual chest needs to be interacted with (which relays fix).

All I'm saying is that shopchests could be made more secure, if they're made more secure shop owners will put more wealth in them (and sink more diamond wealth into building them), and the relative timecost to potential raiding loot stays the same. 

Under current system with one layer of dro protection I wouldn't put more than say a stack of diamonds into a shop chest. If I could have 3 layers of dro protection, it stands to reason I'd put 3 stacks of diamonds into a shop chest. It's the same amount of break time per diamond for a raider.

Protonull:
> Right now the total cost for a MAXIMALLY secure shop chest is 27 diamonds.

Could you diagram this? I don't understand where you're getting that number from.

----

> The contents of the shopchest are broadcasted.

Yes, this is a risk with shops.. and risk is *the whole point.* If we wanted riskless shops, we'd have gone with a digital market with a digital currency. (Yes, I know it's all technically digital, but work with me here.) You willingly take the risk of setting up a shop chest with contents to get the benefit of being able to sell things passively while you're away or offline. You proposal here is simply allowing *you*, or anyone else exceptionally rich (because your whole argument is wealth sink) to almost entirely remove the risks of shop chests. This is why relay recursion was never enabled, because no matter how sinky the wealth sink is, it fundamentally undermines the purpose of ItemExchange introducing risk. If you want to wealth sink protect your shops then be creative and design the most brutally secure shop possible. If you don't want to do that, that's your choice, but *your* decision not to do that doesn't warrant a mechanical change to suit you.

Convoy20:
>Could you diagram this? I don't understand where you're getting that number from.

The design of your 'basic' relay chest - I've already explained why this is the most efficient and best protection.

> Yes, this is a risk with shops.. and risk is _the whole point.

Let's slow this down a little, at what point did I say this removes risk? **The risk / reward is the same** with the only change being a much more exponential cost for shop owners to add more layers.

Protonull:
> The design of your 'basic' relay chest - I've already explained why this is the most efficient and best protection.

And I completely disagree with this assessment.. the third design in the image ("maximum" security relay) goes some way towards showing additional protection. It's not obby-level protection, sure, but not *nothing.*

Convoy20:
> And I completely disagree with this assessment.. the third design in the image ("maximum" security relay) goes some way towards showing additional protection. It's not obby-level protection, sure, but not _nothing._

Which I already explained why it gives you a couple extra minutes of protection, nothing more. 

Do you understand my post above that the risk and reward are the same?

Protonull:
It gives you more than "[just] a couple extra minutes [...], nothing more", yeesh, what a way to spin nearly two hours (for "basic relay design" if using glass).. because remember that's 23.3 minutes *per diamond reinforced glass*, and you can put up to 4 blocks between the shop chest and the ender chest. "Couple extra minutes" what an actual meme.

Convoy20:
> It gives you more than "[just] a couple extra minutes [...], nothing more", yeesh, what a way to spin nearly two hours (for "basic relay design" if using glass).. because remember that's 23.3 minutes _per diamond reinforced glass_, and you can put up to 4 blocks between the shop chest and the ender chest. "Couple extra minutes" what an actual meme.

A couple extra minutes for your 'maximum' design versus the basic relay design, for nearly 2x the cost. Please read through my actual comment before getting angry.

Do you understand my comment above that the risk and reward are the same?

Protonull:
> A couple extra minutes for your 'maximum' design versus the basic relay design, for nearly 2x the cost.

The "maximum" design is an example and is not in any way prescriptive.. you don't *have* to have one layer of obby above the ender chest, and you *can* use any other block other than stone slabs, just be creative. And even then that's assuming swimming/crawling through, which can be mitigated against with a vault bastion since people can't swim/crawl on demand on land without using trap doors or whatever (so on older clients, like 1.12 that the pvpers use, they'd have to mine through the obby facade of the "maximum" security relay). And even then, this is skirting around the fact that it'll still take several hours for someone to break into a suitably bunkerified shop.

wingzero54:
When shop relays were introduced we had a public balance discussion and decided to not allow any additional recursions because they are simply OP. The single ender chest provided a substantial increase in protection for shop chests, but allowing multiple stacked recursions is just broken

PR #232, opened 2021-08-07
Re-Add 2:00 Minute Drinkable Regen

KaloHG:
This was apparently removed post-update. Something that must've been missed in the Humbug update but was never re-added here.

Edit: If people need confirmation that this was a future prior to the update, here's a video https://youtu.be/4uWbINGMeSY at 2:47 shows functionality as intended.

### DIFF
--- a/templates/public/plugins/Finale/config.yml.j2
+++ b/templates/public/plugins/Finale/config.yml.j2
@@ -200,6 +200,13 @@ potions:
 #All splash pots reduced
       splash: true
       multiplier: 1.0
+#2:00 Minute Drinkable Regen, like pre-update.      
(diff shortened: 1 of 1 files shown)


wingzero54:
I have no idea what the context around this is, can you elaborate?

KaloHG:
@wingzero54 in the case of potion buffs this makes drinkable regeneration pots better as they will last longer than a splash pot. This was just the former configuration pre-update to 1.14 as well as the config on 2.0. 

Overall its just re-adding former functionality. 

This is intended for combat & stuff. Making it a bit easier to tank/kite but not overly so if you're stupid.

PR #233, opened 2021-08-10
Add lapis back to enchanting

Convoy20:
Add lapis back to enchanting - don't know why we're still clutching our pearls over lapis being added to enchanting tables back in 1.8 which happened over 6 years ago.


wingzero54:
This is actually something I've been thinking about for a while, and gives lapis more of a need (currently bastions are the only big thing lapis is useful for)

Protonull:
+1

PR #234, opened 2021-08-19
bugged config fix to fix SAH

walkersgaming:
removed ' in phantom message to fix config
credit Ori


wingzero54:
Didn't see this, pushed the change separately

PR #235, opened 2021-08-20
Separate hack configs into their own files

Protonull:
These are the required config changes, assuming https://github.com/CivClassic/SimpleAdminHacks/pull/87 is merged.


wingzero54:
I can't say I'm a fan of this. It is a problem, but I'm not sure this is the solution we should be pursuing

Convoy20:
> I can't say I'm a fan of this. It is a problem, but I'm not sure this is the solution we should be pursuing

Not a fan of the modulization, or not a fan of splitting up the configs into a bunch of different parts?

PR #236, opened 2021-08-21
Removing the uncarve recipes

Lodish:
Removed all the uncarve recipes for converting carved pumpkins into normal pumpkins.


wingzero54:
Thank you

PR #238, opened 2021-08-23
Rollback piston changes, it didn't work

Convoy20:
Adding the piston changes made it so pistons could no longer move observers, totally breaking flying machines and destroying redstone applications.


wingzero54:
You're the one that suggested it :suckfanta:

PR #240, opened 2021-09-02
Concrete Factory Rebalance

Ameliorate:
Concrete powder now takes 64 stone, 1 egg, and 4 dye, bringing it into the scale of materials used by recipes in the aesthetic factory.

There are now harden concrete recipes that convert concrete powder into concrete using 1 water bucket (refunding the bucket)

These changes are untested. I was about to go to sleep, but I did this instead. Errors are expected.

### DIFF
--- a/templates/public/plugins/FactoryMod/config.yml.j2
+++ b/templates/public/plugins/FactoryMod/config.yml.j2
@@ -2710,12 +2710,12 @@ recipes:
     name: Mix White Concrete
     type: PRODUCTION
     input:
-      stone:
-        material: STONE
-        amount: 64
-      egg:
-        material: EGG
-        amount: 1
+      sand:
+        material: SAND
+        amount: 32
+      gravel:
+        material: GRAVEL
+        amount: 32
       white_dye:
         material: WHITE_DYE
         amount: 4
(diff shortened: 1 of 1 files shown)


wingzero54:
I was thinking more like 16 sand/gravel making 64 concrete or something like that. In general the factories are better-than-vanilla but still mimic vanilla recipes, and I like to maintain that rather than departing to cheaper materials

Ameliorate:
Okay. The new version takes 18 sand and 24 gravel, which returns the cobble cost and ratio of materials to pre-grinding mill rebalance.

wingzero54:
Also what about dye? I know dye isn't super expensive but it can be hard to come by in some cases. Would it make sense to lower dye costs

AhriVG:
4 dye per stack is already half the vanilla cost, and in line with the wool, clay, and glass dying factories

Lodish:
Concrete powder being a bit cheaper is fine. Not a fan of having harden recipes, or the recipe ratio being based on the grinding mill.

wingzero54:
> 
> 
> Concrete powder being a bit cheaper is fine. Not a fan of having harden recipes, or the recipe ratio being based on the grinding mill.

Do you have a suggestion for a different ratio? I would imagine it would end up pretty similar

Ameliorate:
I think he probably wants 1:1 sand:gravel, like the vanilla recipe.

wingzero54:
Production time on harden recipes should be 4s, and let's do 16/16/4 sand/gravel/dye for the concrete recipes

Metriximor:
I think wings take on this is perfect

PR #241, opened 2021-09-03
Obfuscate Terracotta to protect Mesa Biomes

Inframission:
Mesa cliffs aren't getting obfuscated as they aren't made of stone like normal terrain


Gjum:
Will this make all terracotta buildings appear as stone from far away?

Inframission:
No this shouldn't visually affect terracotta from the outside. This will make the inside of terracotta obfuscate into ores and air like we're doing to other terrain

PR #242, opened 2021-09-06
Modify Bungee RAM

KaloHG:
Last time we had lag similar to this it turned out to be due to a lack of bungee having RAM, considering the increase in players (from 40-50, to 70-80) it could be possible that 2GB may no longer suffice? Why don't we just give it a shot to see if it might fix some of the network lag issues that players have been experiencing.

### DIFF
--- a/variables/all.yml
+++ b/variables/all.yml
@@ -27,10 +27,10 @@ mc_port: "{{ internal_port }}"
 bungee_port: "{{ external_port }}"
 
 #How much memory is used by the server
-shard_mem: 51G
+shard_mem: 49G
 
 #How much memory is used by bungee
-bungee_mem: 2048M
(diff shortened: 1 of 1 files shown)


KaloHG:
side note: just noticed the TPS seemingly drops when the server hits 70 players, imo bungee highly suspect in this case for the lag 😢

caucow:
> side note: just noticed the TPS seemingly drops when the server hits 70 players, imo bungee highly suspect in this case for the lag 😢

Unless bungee is cutting into the main server's main thread's cpu time, TPS is server side, bungee is a separate service altogether, the two are independent.

KaloHG:
closed since wing resolved it with viaversion changes.

PR #243, opened 2021-09-10
Vary PVP sounds (MASSIVE improvement)

dustcloud777:
First PVP sound pull request was so good, there's another one

### DIFF
--- a/templates/public/plugins/Finale/config.yml.j2
+++ b/templates/public/plugins/Finale/config.yml.j2
@@ -17,9 +17,9 @@ cleanerCombat:
   knockbackSwordsEnabled: true # should knockback swords work?
   sounds:
     weak: false
-    strong: false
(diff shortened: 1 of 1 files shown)


jaymsDooku:
Only turn on crit sounds; do not turn on the other hit sounds please.

wingzero54:
What's your reasoning for wanting to only do some sounds and not others Jayms?

jaymsDooku:
With our custom pvp, the other sounds don't have any significance; there is no sweeping attack, there is no strong attack, there are no weak attacks - there are only crits and normal hits. Therefore, if you're going to re-enable any sound, it should only be the crit sound - do not re-enable the other sounds; they are annoying and have no relationship to actual combat.

Diet-Cola:
Sounds like a /config option sort of thing

PR #252, opened 2021-11-22
Fix damage values

jaymsDooku:
Bring crits back to what it should be, and slightly tone down the overall damage.


dustcloud777:
I do not like this pull req

KaloHG:
wh-what do you mean we can't have critcraft 😭

dustcloud777:
> wh-what do you mean we can't have critcraft 😭

To the contrary, we need to increase crit damage

